PR #6, opened 2019-12-10
adding secret info link in logs

SumiranAgg:


action.yml
```diff
@@ -1,6 +1,6 @@
 # Login to Azure subscription
 name: 'Azure Login'
-description: 'Azure Login wraps the az login, allowing Azure actions to log into Azure or to run Az CLI scripts.'
+description: 'Azure Login wraps the az login, allowing Azure actions to log into Azure or to run Az CLI scripts. github.com/Azure/Actions'
```
niadak: Why this change?

SumiranAgg: Request by @N-Usha so we can show more details about our action in marketplace

PR #14, opened 2020-03-04
Added changes for azure powershell login

aksm-ms:
@SumiranAgg Please review


src/loginAzurePowerShell.ts
```diff
@@ -0,0 +1,76 @@
+import * as os from 'os';
+import * as core from '@actions/core';
+import * as exec from '@actions/exec';
+import * as io from '@actions/io';
+
+var psPath: string;
+
+export const initializeAz = async (servicePrincipalId: string, servicePrincipalKey: string, tenantId: string, subscriptionId: string) => {
```
arjgupta: Do not use Anonymous function. use named functions. Export statements should be confined towards the end of the file.

src/loginAzurePowerShell.ts
```diff
@@ -0,0 +1,76 @@
+import * as os from 'os';
+import * as core from '@actions/core';
+import * as exec from '@actions/exec';
+import * as io from '@actions/io';
+
+var psPath: string;
```
arjgupta: use 'let' instead of 'var'

src/loginAzurePowerShell.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export const initializeAz = async (servicePrincipalId: string, servicePrincipalKey: string, tenantId: string, subscriptionId: string) => {
+    psPath = await io.which("pwsh", true);
+    await importModule();
+    await loginToAzure(servicePrincipalId, servicePrincipalKey, tenantId, subscriptionId);
+}
+
+async function importModule() {
```
arjgupta: Add doc comments for every function. Be precise as to what an input means and what the expected output is.

arjgupta: https://github.com/Microsoft/tsdoc

src/loginAzurePowerShell.ts
```diff
@@ -0,0 +1,76 @@
+import * as os from 'os';
```
arjgupta: Use TSLint.

src/loginAzurePowerShell.ts
```diff
@@ -0,0 +1,76 @@
...
+import * as core from '@actions/core';
+import * as exec from '@actions/exec';
+import * as io from '@actions/io';
+
+var psPath: string;
+
+export const initializeAz = async (servicePrincipalId: string, servicePrincipalKey: string, tenantId: string, subscriptionId: string) => {
+    psPath = await io.which("pwsh", true);
```
asranja: Here we are doing explicit check for pwsh command. If pwsh doesn't exist, the action will throw and login will fail. For non-powershell based actions, it is not really required to login into Azure PowerShell. Logging into Azure CLI should be enough. We should check if we want to explicitly check for pwsh. 

Refer: https://github.com/actions/toolkit/blob/5859d7172e5f4281198f6d74eeb9fe96fb27a23d/packages/io/src/io.ts#L175

src/loginAzurePowerShell.ts
```diff
@@ -0,0 +1,76 @@
...
+        case "Windows":
+        case "Windows_NT":
+            modulePath = `C:\\Modules\\${azPSVersion};`;
+            break;
+        case "macOS":
+        case "Darwin":
+            // TODO: add modulepath
+            break;
+    }
+    process.env.PSModulePath = `${modulePath}${process.env.PSModulePath}`;
```
arjgupta: process.env.PSModulePath will not be available in javascript process.

src/loginAzurePowerShell.ts
```diff
@@ -0,0 +1,76 @@
...
+    const moduleName: string = "Az.Accounts";
+    const azLatestVersion: string = await getLatestModule(moduleName);
+    core.debug(`Az Module version used: ${azLatestVersion}`);
+    setPSModulePath(`${prefix}${azLatestVersion}`);
+}
+
+function setPSModulePath(azPSVersion: string = "") {
+    let modulePath: string = "";
+    const RUNNER: string = process.env.RUNNER_OS || os.type();
+    switch (RUNNER) {
```
asranja: can RUNNER be case sensitive ?

src/loginAzurePowerShell.ts
```diff
@@ -0,0 +1,76 @@
...
+                (New-Object System.Management.Automation.PSCredential('${servicePrincipalId}',(ConvertTo-SecureString ${servicePrincipalKey} -AsPlainText -Force))) \
+                    -Environment ${environment}`);
+    await executePowerShellCommand(`Set-AzContext -SubscriptionId ${subscriptionId} -TenantId ${tenantId}`);
+    await executePowerShellCommand(`Get-AzContext`);
+}
+
+async function executePowerShellCommand(command: string, options: any = {}) {
+    try {
+        await exec.exec(`"${psPath}" -Command "${command}"`, [], options);
+    } catch(error) {
```
asranja: move to next line

src/loginAzurePowerShell.ts
```diff
@@ -0,0 +1,76 @@
...
+        listeners: {
+            stdout: (data: Buffer) => {
+                output += data.toString();
+            },
+            stderr: (data: Buffer) => {
+                error += data.toString();
+            }
+        }
+    };
+    await executePowerShellCommand(`(Get-Module -Name ${moduleName} -ListAvailable | Sort-Object Version -Descending | Select-Object -First 1).Version.ToString()`, options);
```
asranja: leave a line after closing braces

src/loginAzurePowerShell.ts
asranja: leave a line after closing braces

src/loginAzurePowerShell.ts
```diff
@@ -0,0 +1,76 @@
...
+async function importModule() {
+    setPSModulePath();
+    const prefix = "az_";
+    const moduleName: string = "Az.Accounts";
+    const azLatestVersion: string = await getLatestModule(moduleName);
+    core.debug(`Az Module version used: ${azLatestVersion}`);
+    setPSModulePath(`${prefix}${azLatestVersion}`);
+}
+
+function setPSModulePath(azPSVersion: string = "") {
```
arjgupta: add function to validate azPSVersion is a well formed semantic version (<Major>.<Minor>.<Patch>)

src/loginAzurePowerShell.ts
asranja: add some comments as to what this function does. Internally it calls setPSModulePath twice which is not very clear to understand why is it done.

src/loginAzurePowerShell.ts
asranja: call it initializeAzModules ?

src/loginAzurePowerShell.ts
```diff
@@ -0,0 +1,76 @@
...
+            break;
+        case "macOS":
+        case "Darwin":
+            // TODO: add modulepath
+            break;
+    }
+    process.env.PSModulePath = `${modulePath}${process.env.PSModulePath}`;
+}
+
+async function getLatestModule(moduleName: string) {
```
arjgupta: missing return type

src/loginAzurePowerShell.ts
```diff
@@ -0,0 +1,76 @@
...
+            }
+        }
+    };
+    await executePowerShellCommand(`(Get-Module -Name ${moduleName} -ListAvailable | Sort-Object Version -Descending | Select-Object -First 1).Version.ToString()`, options);
+    return output.trim();
+}
+
+async function loginToAzure(servicePrincipalId: string, servicePrincipalKey: string, tenantId: string, subscriptionId: string) {
+    const environment: string = "AzureCloud";
+    await executePowerShellCommand(`Clear-AzContext -Scope Process`);
```
arjgupta: You should put PowerShell scripts being used elsewhere. Maybe in a static class. Or create a ScriptBuilder class. Pass in necessary Parameters and get back the powershell script to be executed.

src/loginAzurePowerShell.ts
```diff
@@ -0,0 +1,76 @@
...
+                    -Environment ${environment}`);
+    await executePowerShellCommand(`Set-AzContext -SubscriptionId ${subscriptionId} -TenantId ${tenantId}`);
+    await executePowerShellCommand(`Get-AzContext`);
+}
+
+async function executePowerShellCommand(command: string, options: any = {}) {
+    try {
+        await exec.exec(`"${psPath}" -Command "${command}"`, [], options);
+    } catch(error) {
+        throw new Error(error);
```
arjgupta: what if there is no catch here ? The Error would be thrown anyway right ? I do not think catch is necessary here.

src/loginAzurePowerShell.ts
arjgupta: I think a better approach would be to define an interface IAzurePowerShellSession . It should have 2 methods: Initialize and Login. Implement a class. we can discuss on this.

src/loginAzurePowerShell.ts
```diff
@@ -0,0 +1,76 @@
...
+    await executePowerShellCommand(`Connect-AzAccount -ServicePrincipal -Tenant ${tenantId} -Credential \
+                (New-Object System.Management.Automation.PSCredential('${servicePrincipalId}',(ConvertTo-SecureString ${servicePrincipalKey} -AsPlainText -Force))) \
+                    -Environment ${environment}`);
+    await executePowerShellCommand(`Set-AzContext -SubscriptionId ${subscriptionId} -TenantId ${tenantId}`);
+    await executePowerShellCommand(`Get-AzContext`);
+}
+
+async function executePowerShellCommand(command: string, options: any = {}) {
+    try {
+        await exec.exec(`"${psPath}" -Command "${command}"`, [], options);
```
arjgupta: you are using pspath here but it is initialized somewhere else. Also, it is a global. Do away with having a global variable pspath and calculate it here. That will be better.

src/loginAzurePowerShell.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export const initializeAz = async (servicePrincipalId: string, servicePrincipalKey: string, tenantId: string, subscriptionId: string) => {
+    psPath = await io.which("pwsh", true);
+    await importModule();
+    await loginToAzure(servicePrincipalId, servicePrincipalKey, tenantId, subscriptionId);
+}
+
+async function importModule() {
+    setPSModulePath();
+    const prefix = "az_";
```
arjgupta: Constants can be put in a separate file.

src/loginAzurePowerShell.ts
```diff
@@ -0,0 +1,76 @@
...
+            modulePath = `/usr/share/${azPSVersion}:`;
+            break;
+        case "Windows":
+        case "Windows_NT":
+            modulePath = `C:\\Modules\\${azPSVersion};`;
+            break;
+        case "macOS":
+        case "Darwin":
+            // TODO: add modulepath
+            break;
```
arjgupta: add default case. throw a message saying unknown os.

src/loginAzurePowerShell.ts
```diff
@@ -0,0 +1,76 @@
...
+    await executePowerShellCommand(`Clear-AzContext -Scope Process`);
+    await executePowerShellCommand(`Clear-AzContext -Scope CurrentUser -Force -ErrorAction SilentlyContinue`);
+    await executePowerShellCommand(`Connect-AzAccount -ServicePrincipal -Tenant ${tenantId} -Credential \
+                (New-Object System.Management.Automation.PSCredential('${servicePrincipalId}',(ConvertTo-SecureString ${servicePrincipalKey} -AsPlainText -Force))) \
+                    -Environment ${environment}`);
+    await executePowerShellCommand(`Set-AzContext -SubscriptionId ${subscriptionId} -TenantId ${tenantId}`);
+    await executePowerShellCommand(`Get-AzContext`);
+}
+
+async function executePowerShellCommand(command: string, options: any = {}) {
```
arjgupta: PowerShell script execution seems like a generic utility. You can take this out as a PowerShellToolRunner and make it available for others to use by placing it in a common area.

src/loginAzurePowerShell.ts
aksm-ms: it is available, i have tested

src/loginAzurePowerShell.ts
aksm-ms: as discussed offline, action will fail if powershell login fails because we do not have input to control it.

src/main.ts
```diff
@@ -25,13 +26,19 @@ async function main() {
...
         if (!servicePrincipalId || !servicePrincipalKey || !tenantId || !subscriptionId) {
             throw new Error("Not all values are present in the creds object. Ensure clientId, clientSecret, tenantId and subscriptionId are supplied.");
         }
-
         await executeAzCliCommand(`login --service-principal -u "${servicePrincipalId}" -p "${servicePrincipalKey}" --tenant "${tenantId}"`);
         await executeAzCliCommand(`account set --subscription "${subscriptionId}"`);
-        console.log("Login successful.");    
+        if (enablePSSession) {
+            console.log(`Running Azure PS Login`);
+            const spnlogin: ServicePrincipalLogin = new ServicePrincipalLogin(servicePrincipalId, servicePrincipalKey, tenantId, subscriptionId);
```
SumiranAgg: why is this a const?

src/main.ts
```diff
@@ -4,13 +4,14 @@ import * as exec from '@actions/exec';
...
+import { ServicePrincipalLogin } from './PowerShell/ServicePrincipalLogin';
 
 var azPath: string;
 var prefix = !!process.env.AZURE_HTTP_USER_AGENT ? `${process.env.AZURE_HTTP_USER_AGENT}` : "";
 
 async function main() {
-    try{
-        // Set user agent varable
+    try {
+        // Set user agent variable
```
SumiranAgg: Can we add comments like:
Attempting Az cli login
Attempting Az PS login
Login via Az cli is successful.
Login via Az pS is successful

src/PowerShell/ServicePrincipalLogin.ts
```diff
@@ -0,0 +1,39 @@
+import * as core from '@actions/core';
+
+import Utils from './Utilities/Utils';
+import PowerShellToolRunner from './Utilities/PowerShellToolRunner';
+import ScriptBuilder from './Utilities/ScriptBuilder';
+import Constants from './Constants';
+
+export class ServicePrincipalLogin implements IAzurePowerShellSession {
+    static readonly environment: string = Constants.environment;
```
SumiranAgg: Rename constants to AzureCloud, Subscription, ServicePrincipal. Looking at the constant, we should understand the context.

src/PowerShell/Utilities/Utils.ts
```diff
@@ -0,0 +1,55 @@
...
+                stdout: (data: Buffer) => {
+                    output += data.toString();
+                },
+                stderr: (data: Buffer) => {
+                    error += data.toString();
+                }
+            }
+        };
+        await PowerShellToolRunner.init();
+        await PowerShellToolRunner.executePowerShellCommand(`(Get-Module -Name ${moduleName} -ListAvailable | Sort-Object Version -Descending | Select-Object -First 1).Version.ToString()`, options);
```
SumiranAgg: As discussed offline with @arjgupta returning output as a structured object makes sense. Error you may handle as you wish.

src/PowerShell/Utilities/Utils.ts
```diff
@@ -0,0 +1,55 @@
...
+            case "linux":
+                modulePath = `/usr/share/${azPSVersion}:`;
+                break;
+            case "windows":
+            case "windows_nt":
+                modulePath = `C:\\Modules\\${azPSVersion};`;
+                break;
+            case "macos":
+            case "darwin":
+                // TODO: add modulepath
```
SumiranAgg: Do we have a module path, if not lets through error.

src/PowerShell/Utilities/ScriptBuilder.ts
```diff
@@ -0,0 +1,16 @@
+export default class ScriptBuilder {
```
SumiranAgg: Should it be a separate class or a function in ServicePrincipalLogin, since it depends on the same state as login object.

src/PowerShell/Utilities/PowerShellToolRunner.ts
```diff
@@ -0,0 +1,16 @@
...
+    static psPath: string;
+
+    static async init() {
+        if(!PowerShellToolRunner.psPath) {
+            PowerShellToolRunner.psPath = await io.which("pwsh", true);
+        }
+    }
+
+    static async executePowerShellCommand(command: string, options: any = {}) {
+        await exec.exec(`"${PowerShellToolRunner.psPath}" -Command "${command}"`, [], options);
```
SumiranAgg: Can you do error handling for az cli login and ps login separately. Bubbling everything to common catch block in main.ts might be a little confusing.

src/main.ts
```diff
@@ -4,13 +4,14 @@ import * as exec from '@actions/exec';
...
 import { FormatType, SecretParser } from 'actions-secret-parser';
+import { ServicePrincipalLogin } from './PowerShell/ServicePrincipalLogin';
 
 var azPath: string;
 var prefix = !!process.env.AZURE_HTTP_USER_AGENT ? `${process.env.AZURE_HTTP_USER_AGENT}` : "";
 
 async function main() {
-    try{
-        // Set user agent varable
+    try {
```
SumiranAgg: Can we refactor az cli login to a different file as well. This will improve readability. Also we can add error handling login individually in each class.

Just read input, preferably in a different function which returns a taskparameter object. Invoke both login classes as required. Wrap within try catch and throw generic error. Rest logic can be handled in individual classes.

src/main.ts
```diff
@@ -4,13 +4,14 @@ import * as exec from '@actions/exec';
 import * as io from '@actions/io';
 
 import { FormatType, SecretParser } from 'actions-secret-parser';
+import { ServicePrincipalLogin } from './PowerShell/ServicePrincipalLogin';
 
 var azPath: string;
 var prefix = !!process.env.AZURE_HTTP_USER_AGENT ? `${process.env.AZURE_HTTP_USER_AGENT}` : "";
```
SumiranAgg: Can you check if the same variable is used as user-agent in powershell. I think the powershell team send some data in an email as for how to add data for telemetry in their user-agent.

action.yml
```diff
@@ -5,6 +5,10 @@ inputs:
   creds:
     description: 'Paste output of `az ad sp create-for-rbac` as value of secret variable: AZURE_CREDENTIALS'
     required: true
+  enable-PSSession:
```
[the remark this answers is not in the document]
N-Usha: Discussed with Akshaya... I propose enable-AzPSSession. Please call out if there are further concerns. Thanks! 

@RoopeshNair FYI

src/PowerShell/Utilities/Utils.ts
aksm-ms: Unable to find modulePath for macOS in task code. Have asked shivangi. Waiting for reply. 
Any other lead on this will be helpful.

src/PowerShell/Utilities/PowerShellToolRunner.ts
aksm-ms: why have a separate catch block, since this catch block just logs the error? This is the same behavior we need in case of Azure PowerShell Login also right? If there is a specific error logic for both logins, maybe we can separate it out.

src/main.ts
aksm-ms: This seems like a good approach, but i think its not within this scope. This can be done later and separately.

src/main.ts
aksm-ms: TSLint suggestion

action.yml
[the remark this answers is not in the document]
chshrikh: can we avoid PSSession wording as this is already a concept in itself, how about enable-AzurePSLogin ?

action.yml
```diff
@@ -5,6 +5,10 @@ inputs:
   creds:
```
arjgupta: The description: "Azure login wraps the az login ..." needs to be updated to specify Azure PowerShell login capability.

action.yml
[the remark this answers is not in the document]
RoopeshNair: +1 on @N-Usha  suggestion for `enable-AzPSSession`

src/main.ts
```diff
@@ -25,13 +29,21 @@ async function main() {
         let servicePrincipalKey = secrets.getSecret("$.clientSecret", true);
         let tenantId = secrets.getSecret("$.tenantId", false);
         let subscriptionId = secrets.getSecret("$.subscriptionId", false);
+        const enablePSSession = core.getInput('enable-PSSession').toLowerCase() === "true";
```
arjgupta: input here should be renamed if that is the case.

src/PowerShell/Utilities/Utils.ts
```diff
@@ -0,0 +1,55 @@
...
+            case "windows_nt":
+                modulePath = `C:\\Modules\\${azPSVersion};`;
+                break;
+            case "macos":
+            case "darwin":
+                throw new Error(`OS not supported`);
+            default:
+                throw new Error(`Unknown os: ${runner.toLowerCase()}`);
+        }
+        process.env.PSModulePath = `${modulePath}${process.env.PSModulePath}`;
```
arjgupta: In linux, if you install PowerShell core, does it populate PSModulePath environment variable even in non-PowerShell based environments. For eg. if i install powershell core on linux and then start a node process, is it guaranteed that there will be an environment variable called PSModulePath in that nodejs process ??

arjgupta: Would be good to log the PSModulePath here.

arjgupta: When this function is called the first time, the value of the azPSVersion param is empty. And so the value of the PSModulePath variable will be the same if the function isn't called. Why do it ?

src/PowerShell/Utilities/Utils.ts
```diff
@@ -0,0 +1,55 @@
...
+            listeners: {
+                stdout: (data: Buffer) => {
+                    output += data.toString();
+                }
+            }
+        };
+        await PowerShellToolRunner.init();
+        await PowerShellToolRunner.executePowerShellScriptBlock(new ScriptBuilder()
+                                .getLatestModuleScript(moduleName), options);
+        const outputJson = JSON.parse(output.trim());
```
arjgupta: nit. JSON means the string notation. at this point you have parsed the output. You can just call it output i believe :D

src/PowerShell/ServicePrincipalLogin.ts
```diff
@@ -0,0 +1,56 @@
...
+            environment: ServicePrincipalLogin.environment,
+            scopeLevel: ServicePrincipalLogin.scopeLevel
+        }
+        const script: string = new ScriptBuilder().getAzPSLoginScript(ServicePrincipalLogin.scheme, this.tenantId, args);
+        await PowerShellToolRunner.init();
+        await PowerShellToolRunner.executePowerShellScriptBlock(script, options);
+        const outputJson: any = JSON.parse(output.trim());
+        if (!(Constants.Success in outputJson)) {
+            throw new Error(`Azure PowerShell login failed with error: ${outputJson[Constants.Error]}`);
+        }
```
arjgupta: Should we add a message here that says "Azure PowerShell session successfully initialized" or something ? Can you check with @SumiranAgg

src/main.ts
aksm-ms: will rename once its confirmed

src/PowerShell/Utilities/Utils.ts
aksm-ms: In case of linux, Az module is present inside folder "/usr/share/". When I run a "Get-Module" command later in the script to get latest version of Az module, Get-Module says "module doesnt exist" because "/usr/share" is not present in the path by default. Hence based on runner, am adding the path which contains the Az module.

arjgupta: oh ok. hmm... ok so from reading the code it isn't obvious why we are doing it. can you add a comment explaining this. Add it where the function is called the first time

package.json
```diff
@@ -17,6 +17,7 @@
     "@actions/core": "^1.1.3",
     "@actions/exec": "^1.0.1",
     "@actions/io": "^1.0.1",
-    "actions-secret-parser": "^1.0.2"
+    "actions-secret-parser": "^1.0.2",
+    "tslint": "^6.1.0"
```
SumiranAgg: tslint should be a devdependency

src/PowerShell/Utilities/PowerShellToolRunner.ts
SumiranAgg: I think having explicit messages like : failed creating az cli session or az ps session, gives customer more insight into what failed. Hence they would try local scripts before they open up a issue.

aksm-ms: In that case we need to use custom exceptions for each login to figure out which failed.
Please refer - https://stackoverflow.com/questions/3239906/multiple-try-catch-or-one

src/PowerShell/ServicePrincipalLogin.ts
SumiranAgg: I think we should. Have similar comment below.

src/PowerShell/Utilities/PowerShellToolRunner.ts
arjgupta: set a flag after az cli succeeds. Use the value of that variable to differentiate between az cli failure and az ps failure.

Issue #17, opened 2020-04-13
Output log reveal azure secrets


aaltotsky:
The usage of the command:
    - uses: azure/login@v1
      with:
        creds: ${{ secrets.AZURE_CREDENTIALS }}      
will output the azure subscription, homeTenantId, and so on into the GitHub Action log. Is there a way to add an additional parameter "--output none"?

aaltotsky:
Perfect! I like the default setting for this. The issue is fixed. Thanks for the quick turnaround on this.

PR #18, opened 2020-04-14
Making login calls silent

ds-ms:
Fixes: #17


ds-ms:
I've done my basic testing. 
This PR is in response to a sev2 incident. I'm going ahead and merging this. 
cc: @arjgupta @SumiranAgg You guys can take a call on what needs to be done here later.

PR #41, opened 2020-07-21
login action for azurestack

rakku-ms:


rakku-ms:
> These changes look good to me. I would let Akshaya or some engineering folks to review as well. Few comments/queries from my side -
> 
> * The new attributes expected in the SPN are optional rt? I mean for non-Azure stack scenarios, user would not need to supply these in the JSON object. Is that rt?

Yes, they are optional. not needed in the JSON object for non-Azure stack scenarios

> * What is the attribute of profileVersion? Are there any acceptable values for it?

Can be one of these {2017-03-09-profile, 2018-03-01-hybrid, 2019-03-01-hybrid, latest}. Azure Stack currently supports '2019-03-01-hybrid'

> * What are the acceptable values for "customEnvironmentName" ? Does the user need to specify in any specific format? 

Can be any unique name (such as AzureStack) other than known clouds ('AzureCloud', 'AzureUSGovernment', etc)

> I assume these additional attributes will be returned when user runs the Az CLI command "az ad sp create-for-rbac --sdk-auth". In case user manually tries to construct the JSOn object, we may need to provide the required guidance for this. Also should we have any additional checks in the code to validate that the values we get for this string follow the required notation?

No, these 2 additional attributes (profileVersion and customEnvironmentName) will not be generated by the above Azure CLI cmd. User will need to manually add these to the existing JSON object. We will provide the required guidance in the doc. We can add the validation for profileVersion parameter but the list of acceptable values will change in future. Hence, we will need to update the code whenever new profileVersion is released. I prefer documenting this instead of adding validation. But based on the review suggestions, I can add validation.

N-Usha:
Thanks Rakesh @rakku-ms . Few additional comments on below

> "No, these 2 additional attributes (profileVersion and customEnvironmentName) will not be generated by the above Azure CLI cmd. User will need to manually add these to the existing JSON object. We will provide the required guidance in the doc. We can add the validation for profileVersion parameter but the list of acceptable values will change in future. Hence, we will need to update the code whenever new profileVersion is released. I prefer documenting this instead of adding validation. But based on the review suggestions, I can add validation."


- If the additional attributes are to be provided manually by the user, we believe that these should be inputs in action.yml than fudging the details into the JSON object. Let the JSON be what the output of Az CLI command is.
- For "customEnvironmentName", we believe there should be a standard list of allowed values which should include one for "AzureStack" which should be validatable in the code and hence its not recommended to accommodate "Can be any unique name other than known clouds "
- IMO seeking "profileVersion" from the customers is not an optimal experience as they may have a dependency on referring to documentation each time for a supported notation and latest version detail. Could we instead avoid this input, hardcode the value and keep bumping it up whenever there is a new API version update? Could you please confirm the release cycle for these versions? How often are they updated? And would there be backward compatibility issues if the version is not updated?

bganapa:
- Agree. instead of placing the details in the json object, it makes sense for them to be part of action.yml, Can you pleas confirm that the values in the action.yml could be overrriden in the individual workflow yml? if the customer want to run a work flow against  azure and azurestack, allowing the override would be helpful.

- customEnvironmentName can be anything, it is not restricted to a known set. For example the customer may have multiple azurestack environments and name them in any way they want. We do not restrict the name to a set in CLI or powershell (az could register -n / Add-AzEnvironment -Name )

- profiles point to a set of RP api versions and allow customers to stick to them without consuming any breaking changes inadvertently.  AzureStack is a boxed product and we support last three versions. These three versions may be on a different profile. There is no ARM call to get to know the supported API profile from the stack environment at the moment. So it is essential to provide the option of  specifying the profile as a parameter. We can have the default value as the latest supported API profile .

src/main.ts
```diff
@@ -30,18 +30,36 @@ async function main() {
...
+        let customEnvironmentName = secrets.getSecret("$.customEnvironmentName", false);
         const enableAzPSSession = core.getInput('enable-AzPSSession').toLowerCase() === "true";
         if (!servicePrincipalId || !servicePrincipalKey || !tenantId || !subscriptionId) {
             throw new Error("Not all values are present in the creds object. Ensure clientId, clientSecret, tenantId and subscriptionId are supplied.");
         }
         // Attempting Az cli login
+        if (customEnvironmentName != "") {
+            console.log(`Registering custom cloud: "${customEnvironmentName}" with ARM endpoint: "${resourceManagerEndpointUrl}"`);
+            try {
+                await executeAzCliCommand(`cloud register -n "${customEnvironmentName}" --endpoint-resource-manager "${resourceManagerEndpointUrl}" `, false);
```
bganapa, line 44: Kevault endpoint suffix and storage endpoint suffix are needed to be passed in, can we make them as parameters as well?

bganapa: or KV and storageendpoints are not needed for the operations being performed here?

rakku-ms: added keyvault and storage endpoints suffixes during cloud registration

rakku-ms:
* Moved profileVersion and customEnvironmentName parameters from secret json object to workflow yml 
* Added keyvault and storage endpoints suffixes during cloud registration

FYI: @N-Usha @bganapa

asranja:
> No, these 2 additional attributes (profileVersion and customEnvironmentName) will not be generated by the above Azure CLI cmd. User will need to manually add these to the existing JSON object. We will provide the required guidance in the doc. We can add the validation for profileVersion parameter but the list of acceptable values will change in future. Hence, we will need to update the code whenever new profileVersion is released. I prefer documenting this instead of adding validation. But based on the review suggestions, I can add validation.

@rakku-ms Will the `az ad sp create-for-rbac` command populate appropriate value for `resourceManagerEndpointUrl` when run against Azure Stack environment or do users need to specify it manually ?

src/main.ts
```diff
@@ -30,18 +30,40 @@ async function main() {
...
         let subscriptionId = secrets.getSecret("$.subscriptionId", false);
+        let resourceManagerEndpointUrl = secrets.getSecret("$.resourceManagerEndpointUrl", false);
+        let profileVersion = core.getInput("profileVersion");
+        let customEnvironmentName = core.getInput("customEnvironmentName");
         const enableAzPSSession = core.getInput('enable-AzPSSession').toLowerCase() === "true";
         if (!servicePrincipalId || !servicePrincipalKey || !tenantId || !subscriptionId) {
             throw new Error("Not all values are present in the creds object. Ensure clientId, clientSecret, tenantId and subscriptionId are supplied.");
         }
         // Attempting Az cli login
+        if (customEnvironmentName != "") {
```
asranja, line 41: Can we fix the customEnvironmentName to be AzureStack and check for that condition here?

src/main.ts
```diff
@@ -30,18 +30,40 @@ async function main() {
         let servicePrincipalKey = secrets.getSecret("$.clientSecret", true);
         let tenantId = secrets.getSecret("$.tenantId", false);
         let subscriptionId = secrets.getSecret("$.subscriptionId", false);
+        let resourceManagerEndpointUrl = secrets.getSecret("$.resourceManagerEndpointUrl", false);
+        let profileVersion = core.getInput("profileVersion");
```
asranja, line 34: inputs have to be added in action.yml file as well

src/main.ts
```diff
@@ -30,18 +30,40 @@ async function main() {
         let servicePrincipalKey = secrets.getSecret("$.clientSecret", true);
         let tenantId = secrets.getSecret("$.tenantId", false);
         let subscriptionId = secrets.getSecret("$.subscriptionId", false);
+        let resourceManagerEndpointUrl = secrets.getSecret("$.resourceManagerEndpointUrl", false);
+        let profileVersion = core.getInput("profileVersion");
+        let customEnvironmentName = core.getInput("customEnvironmentName");
```
asranja, line 35: call it just environment instead of  customEnvironmentName?

src/main.ts
```diff
@@ -30,18 +30,40 @@ async function main() {
...
             throw new Error("Not all values are present in the creds object. Ensure clientId, clientSecret, tenantId and subscriptionId are supplied.");
         }
         // Attempting Az cli login
+        if (customEnvironmentName != "") {
+            if (!resourceManagerEndpointUrl) {
+                throw new Error("resourceManagerEndpointUrl is a required parameter when customEnvironmentName is defined.")
+            }
+            console.log(`Registering custom cloud: "${customEnvironmentName}" with ARM endpoint: "${resourceManagerEndpointUrl}"`);
+            try {
+                let suffixKeyvault = ".vault" + resourceManagerEndpointUrl.substring(resourceManagerEndpointUrl.indexOf('.'))
```
asranja, line 47: missing semi-colon at end. TS ignores it but it will be good to be consistent with the rest of the code.

src/main.ts
N-Usha: @rakku-ms - Could you please confirm if there is a strong requirement for knowing the Environment "name"? Or could we just seek Environment "type" which can just be an allowed list of keywords like "AzureStack" ?

bganapa: yes, A customer can have several instances of AzureStack and the name cannot be limited to a particular set.  You can see that when we do 'az cloud register' , the name is a parameter for customer to name it whatever they want.

src/main.ts
bganapa: environmentName sounds good to me

rakku-ms:
> > No, these 2 additional attributes (profileVersion and customEnvironmentName) will not be generated by the above Azure CLI cmd. User will need to manually add these to the existing JSON object. We will provide the required guidance in the doc. We can add the validation for profileVersion parameter but the list of acceptable values will change in future. Hence, we will need to update the code whenever new profileVersion is released. I prefer documenting this instead of adding validation. But based on the review suggestions, I can add validation.
> 
> @rakku-ms Will the `az ad sp create-for-rbac` command populate appropriate value for `resourceManagerEndpointUrl` when run against Azure Stack environment or do users need to specify it manually ?

Yes, sp Azure CLI cmd will populate the correct value for `resourceManagerEndpointUrl` when run against AAD based Azure Stack. We don't support this cmd for ADFS based Azure Stack. We plan to write a json generator script to do this job for ADFS based Azure Stack.

src/main.ts
rakku-ms: ok.

src/main.ts
rakku-ms: will change to environmentName

src/main.ts
rakku-ms: Yes, thanks for catching this!

rakku-ms: updated

src/main.ts
asranja: @bganapa  
- In our Azure DevOps world, we use a list of pre-defined values for the different cloud types. Users targeting Azure Stack, only needs to select AzureStack as the environment and provide the Azure Stack server URL.  

89559574-7836a400-d833-11ea-9d64-8bcdafcdbfa1

- The `az cloud register` persists for only one session and cannot be reused in a new session. Also, a user cannot login into multiple environments in a single workflow job. I am not sure if we are gaining anything by letting a user name the environment whatever they want. If the user wants to differentiate between the instances, he can specify it in the job name, something like :
```
name: Run on Azure Stack
jobs: 
  run-on-instance-1:
   - uses: azure/login@v1
        with:
          environment: AzureStack
          creds: ${{ secrets.AZURE_CREDENTIALS_FOR_INSTANCE_1 }}

   run-on-instance-2:
   - uses: azure/login@v1
        with:
          environment: AzureStack
          creds: ${{ secrets.AZURE_CREDENTIALS_FOR_INSTANCE_2 }}
```
- We also do not ask for profile version in Azure DevOps, just the Server Url. Can this value be fixed in the action instead of taking input from user.
 
Capture

rakku-ms: @asranja I am using self-hosted runner as Azure Stack env is not publicly accessible from internet. For me registered cloud is getting persisted on my desktop. So, I am unable to run workflow on 2 different Azure Stack environment with same environmentName `AzureStack`. The second run also happens against the first since the cloud with the name `AzureStack` already exists. The register cmd doesn't overwrite the previous environment and just try to login to the existing one. One workaround would be to first remove the already existing `AzureStack` environment and then try to register. If this is what your team prefers, I will make this change. FYI: @bganapa  
```
PS login> az cloud list --output table
IsActive    Name                  Profile
----------  --------------------  -----------------
False       AzureCloud            latest
False       AzureChinaCloud       latest
False       AzureUSGovernment     latest
False       AzureGermanCloud      latest
False       AzureStackRedmond001  2019-03-01-hybrid
False       AzureStackTest01      2019-03-01-hybrid
False       AzureStackTest02      2019-03-01-hybrid
False       AzureStackLH01        2019-03-01-hybrid
False       AzureStackRedmond     2019-03-01-hybrid
```

asranja: @rakku-ms Thanks for giving more details regarding the environment name. We will discuss this internally and get back.

Regarding the profile version: As per my understanding, most Azure resources will have different api-version depending on the profile version. Currently, we have hard-coded all the api-versions in our REST Apis, for [example](https://github.com/Azure/sql-action/blob/71cbe30f683e929e6b48173468b877fb22e66009/src/AzureSqlResourceManager.ts#L114). I am not sure how are we planning on updating the rest api-versions depending on the profile version.

N-Usha: @rakku-ms - Can you please help understand how many customers usually need to access two or more different Azure stack environments?

src/PowerShell/Utilities/ScriptBuilder.ts
```diff
@@ -9,6 +9,9 @@ export default class ScriptBuilder {
         let command = `Clear-AzContext -Scope Process;
              Clear-AzContext -Scope CurrentUser -Force -ErrorAction SilentlyContinue;`;
         if (scheme === Constants.ServicePrincipal) {
+            if (args.environment == "AzureStack") {
+                command += `Add-AzEnvironment -Name ${args.environment} -ARMEndpoint ${args.resourceManagerEndpointUrl} | out-null;`;
```
asranja, line 13: Do we not need keyvault and storage suffixes as in the case of Az CLI ?

src/main.ts
```diff
@@ -30,18 +30,45 @@ async function main() {
...
             throw new Error("Not all values are present in the creds object. Ensure clientId, clientSecret, tenantId and subscriptionId are supplied.");
         }
         // Attempting Az cli login
+        if (environment == "AzureStack") {
+            if (!resourceManagerEndpointUrl) {
+                throw new Error("resourceManagerEndpointUrl is a required parameter when environment is defined.");
+            }
+            console.log(`Unregistering cloud: "${environment}" first if it exists`);
+            try {
+                await executeAzCliCommand(`cloud set -n AzureCloud`, true);
```
asranja, line 46: is this required ? are you doing this to ensure that your current cloud is not set as AzureStack before unregistering it ?

rakku-ms: Yes, unregister cmd fails if the current cloud is set to AzureStack. So, I am switching the active cloud to default (AzureCloud) before unregistering AzureStack.

src/PowerShell/Utilities/ScriptBuilder.ts
bganapa: if we are not doing any storage/keyvault data operations , we dont need them

rakku-ms: We don't want to put any restriction on the Az powershell script, so it's better if I set it here as well.

rakku-ms: we don't need to set storage/keyvault suffixes for Az PowerShell here. these are already set correctly by default when -ARMEndpoint parameter is passed.

rakku-ms:
@asranja let us know if you have any other concerns regarding this PR. If not, can we please publish this?
FYI, @bganapa

action.yml
```diff
@@ -9,6 +9,10 @@ inputs:
     description: 'Set this value to true to enable Azure PowerShell Login in addition to Az CLI login'
     required: false
     default: false
+  environment:
+    description: 'Name of the environment (Set this to AzureStack for AzureStack environment)'
```
asranja, line 13: @N-Usha for the description.

N-Usha: We also need to mention a default value for this parameter with value: public. Description can be "Name of the environment. Supported values are AzureStack, public. Default being public". And is the value expected to be case sensitive?

N-Usha (changes requested):

action.yml
rakku-ms: @N-Usha I have updated the description. Also made modification to make this parameter case insensitive.

N-Usha (approved):

N-Usha:
IMO we should NOT be merging the changes into master but rather into a new branch exclusively for AzureStackSupport-Beta... @nadesu to reconfirm.  I or one of our team members can create the new branch if you want us to.. @rakku-ms -  Please confirm.

nadesu:
> IMO we should NOT be merging the changes into master but rather into a new branch exclusively for AzureStackSupport-Beta... @nadesu to reconfirm. I or one of our team members can create the new branch if you want us to.. @rakku-ms - Please confirm.

I agree with the above proposal

rakku-ms:
> > IMO we should NOT be merging the changes into master but rather into a new branch exclusively for AzureStackSupport-Beta... @nadesu to reconfirm. I or one of our team members can create the new branch if you want us to.. @rakku-ms - Please confirm.
> 
> I agree with the above proposal

@N-Usha @nadesu yes, please create a new branch for AzureStackSupport. FYI, @bganapa @sarathys

N-Usha:
> > > IMO we should NOT be merging the changes into master but rather into a new branch exclusively for AzureStackSupport-Beta... @nadesu to reconfirm. I or one of our team members can create the new branch if you want us to.. @rakku-ms - Please confirm.
> > 
> > 
> > I agree with the above proposal
> 
> @N-Usha @nadesu yes, please create a new branch for AzureStackSupport. FYI, @bganapa @sarathys

https://github.com/Azure/login/tree/AzureStackSupport-Beta - Please use this branch.

asranja (approved):

rakku-ms:
> https://github.com/Azure/login/tree/AzureStackSupport-Beta - Please use this branch.

thanks @N-Usha!
FYI, @garvitarai

PR #51, opened 2020-09-11
silenting az cli version command

aksm-ms:


lib/main.js
```diff
@@ -37,7 +37,7 @@ function main() {
             core.exportVariable('AZURE_HTTP_USER_AGENT', userAgentString);
             core.exportVariable('AZUREPS_HOST_ENVIRONMENT', azurePSHostEnv);
             azPath = yield io.which("az", true);
-            yield executeAzCliCommand("--version");
```
arjgupta, line 40: what happens when passing in 'true' ?

20shivangi (approved):

lib/main.js
arjgupta: don't we need the output of az --version for debugging purposes ?

aksm-ms:
Closing since opened a new PR to fix this.

Issue #20, opened 2020-04-15
Login consistently takes 30s-60s  to complete


ghstahl:
Is this a function of the VMs that github provisions, this action, or the az cli itself?  

```
az login 27s (UP TO 60 seconds)
Login successful.
Run azure/login@v1
C:\windows\system32\cmd.exe /D /S /C ""C:\Program Files (x86)\Microsoft SDKs\Azure\CLI2\wbin\az.cmd" --version"
WARNING: You have 2 updates available. Consider updating your CLI installation. Instructions can be found at https://docs.microsoft.com/en-us/cli/azure/install-azure-cli
azure-cli                          2.2.0 *

command-modules-nspkg              2.0.3
core                               2.2.0 *
nspkg                              3.0.4
telemetry                          1.0.4

Extensions:
azure-devops                      0.17.0

Python location 'C:\Program Files (x86)\Microsoft SDKs\Azure\CLI2\python.exe'
Extensions directory 'C:\Program Files\Common Files\AzureCliExtensionDirectory'

Python (Windows) 3.6.6 (v3.6.6:4cf1f54eb7, Jun 27 2018, 02:47:15) [MSC v.1900 32 bit (Intel)]

Legal docs and information: aka.ms/AzureCliLegal




Please let us know how we are doing: https://aka.ms/clihats
Login successful.
```

eaarora-ms:
@ghstahl I am afraid I do not understand your query. With the Azure login Action, you can automate your workflow to do an Azure login using Azure service principal and run Az CLI and Azure PowerShell scripts. Are you facing any issue in using this action?

ljtill:
Also experiencing this behavior, we're using Azure PowerShell process. Runs can vary from 30 seconds to 4 minutes to complete.

```yaml
- name: Authenticate session
        uses: azure/login@v1.1
        with:
          creds: ${{ secrets.AZURE_CREDENTIALS }}
          enable-AzPSSession: true
```

```text
2020-06-08T12:33:20.0747185Z ##[group]Run azure/login@v1.1
2020-06-08T12:33:20.0747527Z with:
2020-06-08T12:33:20.0748054Z   creds: ***
2020-06-08T12:33:20.0748329Z   enable-AzPSSession: true
2020-06-08T12:33:20.0748820Z env:
2020-06-08T12:33:20.0749734Z   AZURE_CREDENTIALS: ***
[...]

PR #61, opened 2020-10-02
Check Az Powershell version via file system

harvey-k:
Since so many versions of Az Powershell are installed on the virtual environments, Get-Module can be slow. Instead, obtain available Az Powershell versions directly from the filesystem. #20

### DIFF
--- a/__tests__/PowerShell/Utilities/Utils.test.ts
+++ b/__tests__/PowerShell/Utilities/Utils.test.ts
@@ -20,26 +20,26 @@ describe('Testing isValidVersion', () => {
 });
 
 describe('Testing setPSModulePath', () => {
-    test('PSModulePath with azPSVersion non-empty', () => {
-        Utils.setPSModulePath(version);
+    test('PSModulePath with azPSVersion non-empty', async () => {
+        await Utils.setPSModulePath(version);
         expect(process.env.PSModulePath).toContain(version);
     });
-    test('PSModulePath with azPSVersion empty', () => {
+    test('PSModulePath with azPSVersion empty', async () => {
         const prevPSModulePath = process.env.PSModulePath;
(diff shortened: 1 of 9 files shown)


arjgupta:
@harvey-k this approach will not work for private runners. Your changes are specific to hosted runner and private runners will not follow the scheme of storing azure modules in a separate path. Which is why it is better to use Get-Module so as to abstract away this detail

harvey-k:
@arjgupta  

Sorry, I don't follow why this wouldn't work for private runners. The existing Get-Module codepath was kept as a fallback:
https://github.com/Azure/login/pull/61/commits/49ef0a1ea893fe6edcbe21ea8170a48d8c0af95f#diff-6f6a3b4e60ef55a757a45b62b60aa82eb2df252eb4a96cc8601ada2411f87514R50

I understand this will not speed up private runners, but I would imagine they do not experience this issue due to only having a few powershell versions installed.

Do you want a different check for determining if the code is executing on a hosted runner?

aksm-ms:
@harvey-k Thanks for the contribution.

Following are the possible failure cases of this approach:
- **For both Hosted and private runners**: Since folder names are used to find the latest module, there could be a mismatch between folder name and the actual version of the module installed in the folder. This issue does not occur while using `Get-Module` because it reads the version from manifest file of the module.
- **Private runners**: Although the folder az_<x.y.z> might be present inside the root folder, it could have some other contents other than the az module in which case the action fails.
Also this approach is tied to a specific folder path which is unlike PowerShell way of implementation.

Hence, at the moment, these changes will not be taken. We are currently investigating the performance issue on Login action, we will address the issue soon.

cc: @arjgupta

aksm-ms:
Please feel free to open in case of any queries.

Issue #56, opened 2020-09-28
Self-Hosted Github Runner in Azure VM using Managed Identity


jdelforno:
Thanks for your efforts on writing this!

Is it possible to add support for Managed Identities using a self-hosted github runner in an Azure VM / Container?

As the Action already uses Az Module, I'd hope it'd be a simple switch to use MSI ala: https://docs.microsoft.com/en-us/powershell/azure/authenticate-azureps?view=azps-4.7.0#sign-in-using-a-managed-identity

Doing so would remove any potential for secret leakage anywhere along the line.

Cheers,
JD

aksm-ms:
@jdelforno Thanks for the suggestion. I will mark it as enhancement and we will work on it.

mudunuax:
Hi. Can you please share any timeline, if there is, on when this enhancement will be merged to the master?

jdelforno:
> 
> 
> I am also curious about the timeline.
> 
> Workaround, for now, would be something like this:
> 
> ```
>     - name: Azure CLI script
>       uses: azure/CLI@v1
>       with:
>         azcliversion: 2.0.72
>         inlineScript: |
>            az login --identity
> ```
> 
> However, it would look cleaner if there was an option to do:
> 
> ```
>     - name: Login via Az module
>       uses: azure/login@v1
>       with:
>         identity: true
> ```
> 
> where another parameter like `username` is an optional value of <client_id|object_id|resource_id>

That's a great step forward though. Especially with the Github Branch / Environment with secret protection in place.

cmatskas:
Hi all, can someone let me know if this was ever implemented? We have a working solution [here](https://github.com/Azure/login/pull/68) that JP recommended earlier and I would love to freshen up and send a PR if you're open to it?? @aksm-ms @BALAGA-GAYATRI

mikeviviani:
Hi
adding here the interest of this feature. Any plan to release it?

dstratman-fi:
Team, this is sorely needed.  This feature for the azure cli has been available for some time, and this looks like an easy win / implementation, can we get an estimate or timeline?

YanaXu:
This feature is supported by `v1` and `v1.5.1` now.

PR #68, opened 2020-10-12
Adding Azure managed identity support (#56)

jpda:
Adds user-assigned and system-assigned managed identity support (#56) for self-hosted runners.


action.yml
```diff
@@ -4,7 +4,21 @@ description: 'Authenticate to Azure and run your Az CLI or Az
 inputs: 
   creds:
     description: 'Paste output of `az ad sp create-for-rbac` as value of secret variable: AZURE_CREDENTIALS'
-    required: true
+    required: false
```
arjgupta, line 7: Should there instead be an input called 'credentialprovider' or 'loginprovider' which takes values like spnwithsecret, spnwithcertificate, and managedidentity and then the action performs login based on this input value. @N-Usha thoughts ?

jpda:
Going to close this for a bit while I refactor to make it easier to merge

maisoenaingtun:
Support {{#each}}-{{else}}#56

PR #69, opened 2020-10-12
Bump the versions in the readme

Fokko:
Otherwise I get nasty deprecation warnings

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
-    - uses: azure/login@v1.1
+    - uses: azure/login@v1.1.1
       with:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -33,7 +33,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
-    - uses: azure/login@v1.1
+    - uses: azure/login@v1.1.1
```
aksm-ms, line 36: @Fokko while using the action, use the major version so that you will get the latest updated version. Eg: azure/login@v1. Here in documentation, we will update it to point major version.

aksm-ms:
Closing since there is a new PR to fix the readme.

PR #73, opened 2020-10-20
Added no subscription support

Ganeshrockz:
This PR includes the support to allow az login with `--allow-no-subscription` flag. The action will accept a new optional input `allow-no-subscriptions` which when passed as `true` will support az login for SPNs created at a Management group level scope

### DIFF
--- a/src/PowerShell/Utilities/ScriptBuilder.ts
+++ b/src/PowerShell/Utilities/ScriptBuilder.ts
@@ -12,9 +12,7 @@ export default class ScriptBuilder {
             command += `Connect-AzAccount -ServicePrincipal -Tenant '${tenantId}' -Credential \
             (New-Object System.Management.Automation.PSCredential('${args.servicePrincipalId}',(ConvertTo-SecureString '${args.servicePrincipalKey.replace("'", "''")}' -AsPlainText -Force))) \
                 -Environment '${args.environment}' | out-null;`;
-            if (args.scopeLevel === Constants.Subscription && 
-                args.subscriptionId && 
-                args.subscriptionId.length > 0) {
+            if (args.scopeLevel === Constants.Subscription && !args.allowNoSubscriptionsLogin) {
                 command += `Set-AzContext -SubscriptionId '${args.subscriptionId}' -TenantId '${tenantId}' | out-null;`;
             }
         }
--- a/src/PowerShell/Utilities/ScriptBuilder.ts
+++ b/src/PowerShell/Utilities/ScriptBuilder.ts
@@ -12,7 +12,9 @@ export default class ScriptBuilder {
             command += `Connect-AzAccount -ServicePrincipal -Tenant '${tenantId}' -Credential \
             (New-Object System.Management.Automation.PSCredential('${args.servicePrincipalId}',(ConvertTo-SecureString '${args.servicePrincipalKey.replace("'", "''")}' -AsPlainText -Force))) \
                 -Environment '${args.environment}' | out-null;`;
-            if (args.scopeLevel === Constants.Subscription) {
+            if (args.scopeLevel === Constants.Subscription && 
+                args.subscriptionId && 
+                args.subscriptionId.length > 0) {
                 command += `Set-AzContext -SubscriptionId '${args.subscriptionId}' -TenantId '${tenantId}' | out-null;`;
             }
         }
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
-    - uses: azure/login@v1.1
+    - uses: azure/login@v1
       with:
         creds: ${{ secrets.AZURE_CREDENTIALS }}
 
@@ -57,7 +57,7 @@ jobs:
     steps:
 
     - name: Login via Az module
-      uses: azure/login@v1.1
+      uses: azure/login@v1
       with:
         creds: ${{secrets.AZURE_CREDENTIALS}}
         enable-AzPSSession: true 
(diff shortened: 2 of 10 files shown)


ds-ms (approved):
LGTM

ds-ms:
Add an L0 as well

aksm-ms:
> Add an L0 as well

Yes, please add an L0.

Ganeshrockz:
> > Add an L0 as well
> 
> Yes, please add an L0.

Done

Ganeshrockz:
@aksm-ms Can you please review and approve this PR? We have to move this to production quickly to support one of our critical flows.Thanks!

cc: @ammohant @ds-ms

aksm-ms (commented):
Code changes looks good to me. There are many other actions which use the context set by login action. 
@Ganeshrockz  can you send me output of "az account show" after running az login using allow-no-subscriptions?

src/PowerShell/Utilities/ScriptBuilder.ts
```diff
@@ -12,7 +12,9 @@ export default class ScriptBuilder {
             command += `Connect-AzAccount -ServicePrincipal -Tenant '${tenantId}' -Credential \
             (New-Object System.Management.Automation.PSCredential('${args.servicePrincipalId}',(ConvertTo-SecureString '${args.servicePrincipalKey.replace("'", "''")}' -AsPlainText -Force))) \
                 -Environment '${args.environment}' | out-null;`;
-            if (args.scopeLevel === Constants.Subscription) {
+            if (args.scopeLevel === Constants.Subscription && 
+                args.subscriptionId && 
+                args.subscriptionId.length > 0) {
```
asranja, line 17: why is the `args.subscriptionId.length > 0` required ? you could use `!!args.subscriptionId`

ds-ms: Yep. that works too. +1

aksm-ms (commented):
Currently many azure actions like Azure Mysql, Azure Sql-action, Azure Postgresql, Azure Webapps deploy,etc are using the context set by Azure login action.

When a service principal under the scope "subscription" is used, az context id will contain the subscriptionId whereas if a service principal under the scope "management-group" is used, context will contain tenantId.

If we use --allow-no-subscription, it will break the current flow of subsequent actions which are dependent on the subscriptionId. For example, if they try to use Azure Mysql action with azure login with management scope service principal, it will fail.

arjgupta:
@Ganeshrockz , if the service principal in question has access over a subscription (read or higher) in addition to any other access it needs to execute your use case, you would still be able to run the action as is and achieve your use case.

As for this change, if a subscription id isn't passed in and we continue ahead with the login, then a subsequent action like web app deploy or sql deploy will fail (because, after all, we did not specify a subscription at login time). This will make it harder to use most of the typical actions. @N-Usha @aksm-ms and I discussed how we can support such scenarios and came to the conclusion that we will need to come up with a more detailed spec before we can commit to anything here.

ds-ms:
@arjgupta What do you recommend for logging into an azure context with a management scope alone then? 

Given that `subscriptionId` is kind of mandatory in the current state of action, I'll have to do an extra role-binding to some subscription so that the login works for me, right?

aksm-ms:
> I disagree with this, `--allow-no-subscriptions` is not a default. It's conditional based on the inputs, right? If the user consciously wants to use their SPN without binding it to any subscription, we should let them right?

@ds-ms Yes, --allow-no-subscription is not a default at login action level.  Lets take an example of Azure Mysql.
Azure Mysql action currently needs subscriptionId to call some ARM APIs. If a user is able to login using a service principal with management scope and his mysql action fails, it means that we are supporting a backward compatible feature in login but not across the end to end flow for all other azure actions.

I have created a SP at management scope and tried the changes in this PR with Azure Mysql action, it failed. Please find the logs:

**Workflow:**
image

**Logs:**
image

ds-ms:
>Yes, --allow-no-subscription is not a default at login action level. Lets take an example of Azure Mysql.
Azure Mysql action currently needs subscriptionId to call some ARM APIs. If a user is able to login using a service principal with management scope and his mysql action fails, it means that we are supporting a backward compatible feature in login but not across the end to end flow for all other azure actions.

That means the user consciously changed things. The mysql action failed because you, as a user, haven't provided a subscription id in your Azure CREDS, so we didn't even set a subscription context. I feel that this expectation is fair.

aksm-ms:
> That means the user consciously changed things. The mysql action failed because you, as a user, haven't provided a subscription id in your Azure CREDS, so we didn't even set a subscription context. I feel that this expectation is fair.
@N-Usha Please provide your inputs

N-Usha:
> That means the user consciously changed things. The mysql action failed because you, as a user, haven't provided a subscription id in your Azure CREDS, so we didn't even set a subscription context. I feel that this expectation is fair.

User may have consciously or unconsciously omitted the subscription ID in the Creds w/o expecting the consequences that most of the other actions would fail. In most enterprises, SPNs are created by IT admins and are passed on to the developers who may or may not understand the scope of the SPN until they actually use it.  My main concern here is that the error messages thrown by the failing actions are not meaningful or actionable to let the user know how the user could fix the error. And fixing the error messages across all existing actions is simply unscalable.

And hence we believe that we need to make it more explicit an action from user than simply omitting the subscription ID from SPN JSOn object. Potentially by including a new input parameter in the action.yml for login action called "scope" which by default takes "subscription scope". User would then need to explicitly change its value to "Management Scope" and can then omit the subscription ID from SPN. However we also need to understand what is needed for supporting rest of the scopes like RG or resource aswell. Which is why we need to spec this comprehensively before making any changes.

raiyanalam:
As discussed in the meeting today, 
- We will add a new input `allow-no-subscriptions` which will take boolean values. By default the value will be `false`.  Whenever it is `true`, the action will pass this flag while logging into Azure.
- This will be a minor update to the action version.
- @ds-ms @Ganeshrockz to update the PR accordingly.

Ganeshrockz:
@arjgupta / @asranja / @aksm-ms  I have made the changes as discussed. Please review the PR

cc: @ds-ms @raiyanalam

ds-ms (approved):
LGTM

aksm-ms (approved):

aksm-ms:
LGTM. @arjgupta please take a look.

arjgupta (approved):

PR #76, opened 2020-11-03
adding az cli version debug logs

aksm-ms:
#51

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -21,8 +21,17 @@ async function main() {
         core.exportVariable('AZUREPS_HOST_ENVIRONMENT', azurePSHostEnv);
 
         azPath = await io.which("az", true);
-        await executeAzCliCommand("--version");
-
+        let output: string = "";
+        const options: any = {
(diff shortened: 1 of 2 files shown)


src/main.ts
```diff
@@ -59,13 +68,14 @@ async function main() {
     }
 }
 
-async function executeAzCliCommand(command: string, silent?: boolean) {
+async function executeAzCliCommand(command: string, silent?: boolean, options: any = {}) {
+    options.silent = !!silent;
     try {
-        await exec.exec(`"${azPath}" ${command}`, [],  {silent: !!silent}); 
+        await exec.exec(`"${azPath}" ${command}`, [],  options);
```
arjgupta, line 74: When doing az login, will this not show the service principal secret value ?

aksm-ms: no it will not show, because i am adding silent in options in previous line

arjgupta: ok.

arjgupta (approved):

Issue #42, opened 2020-07-26
Symbols in az-cli generated clientSecret breaks login


walzph:
Following [this](https://docs.microsoft.com/en-us/azure/container-instances/container-instances-github-action) guide, most people probably just copy&paste the JSON output generated by the service principal creation via _az-cli_ to their Github secrets.

However the generated _clientSecret_ often contains symbols which must be escaped in order to work in the `az login` command performed in the Github Action.

For example:
_az-cli_ generated a _clientSecret_ which includes `avc!bH9;5<d9L,M!`.  
This breaks the login process without a meaningful error mesasge. It works if it's escaped to `avc\!bH9\;5\<d9L\,M\!` before being inserted here:
```
executeAzCliCommand(`login --service-principal -u "${servicePrincipalId}" -p "${servicePrincipalKey}" --tenant "${tenantId}"`, true);
```

arjgupta:
@walzph thanks for reporting the issue. We will add logic to escape special characters in client secret.

aksm-ms:
@walzph Can you please paste the error logs and also the logs after escaping the secret here?

aksm-ms:
Issue is fixed in PR.

PR #80, opened 2020-11-04
adding params to escape symbols in az cli

aksm-ms:
#42

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -96,10 +96,10 @@ async function main() {
     }
 }
 
-async function executeAzCliCommand(command: string, silent?: boolean, options: any = {}, parameters: any = []) {
-    options.silent = !!silent;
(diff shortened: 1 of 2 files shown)


src/main.ts
```diff
@@ -79,10 +96,10 @@ async function main() {
     }
 }
 
-async function executeAzCliCommand(command: string, silent?: boolean, options: any = {}) {
+async function executeAzCliCommand(command: string, silent?: boolean, options: any = {}, parameters: any = []) {
```
asranja, line 99: having both options and parameters might be confusing. parameters seems like Az cli arguments to me and options refer to execOptions. Can we update this ?

aksm-ms: "parameters" must be renamed to arguments?

aksm-ms: made changes

asranja (approved):

Issue #39, opened 2020-06-29
Feature Request: Ability to Pass Authentication Parameters Individually 


mikaelweave:
It would be great if this action allowed passing in the Subscription ID, Tenant ID, Client ID, and Client Secret as individual parameters instead of bundled in a single JSON object (`creds` today). This would allow for easier use in pipelines that use Terraform (as it requires them split out). It would also make secret rotation easier as you don't have to rebuilt the JSON object.

Potential syntax (as an alternative to the existing syntax):

```yaml
- uses: azure/login@v1.1
      with:
        client_id: ${{ secrets.CLIENT_ID }}
        client_secret: ${{ secrets.CLIENT_SECRET }}
        subscription_id: ${{ secrets.SUBSCRIPTION_ID }}
        tenant_id: ${{ secrets.TENANTID }}
```

If you are open to PRs, I'd love to try this!

ericsampson:
We would really love this too. We have the same 4 separate secret setup as you @mikaelweave , used with Terraform : )
It would be awesome to not need duplicate secrets as JSON to pass into this Action.

kmkumaran:
The reason the login Action took the [credential parameters\(https://github.com/Azure/login/blob/master/README.md#configure-deployment-credentials) as single Json is with the expectation that the getting started experience for users will be simplified with just one copy / paste of command line output as against multiple user actions. 

It appears with Terraform, the need is to have them independent. @N-Usha / @azooinmyluggage / @nadesu - Can you follow-up on this ask and see how we can meet this need and still retain simplified experience for users who would like to opt the current way.

ericsampson:
Thanks @kmkumaran 👍

azooinmyluggage:
One other reason was to use the output of the az ad sp create-for-rbac command as an input. Both arguments at the end are for keeping it simple for end user. 

But looking at the description by @ericsampson and the up votes, this needs to change.

@ericsampson we are open to PR.  How can we solve it gracefully? Two potential guidelines are:
1. Only one set of credentials work. Is there any precedence on one over the other?
2. Do we need any renaming? Having a json based input called "creds" and then four other inputs. Is there anyway we can implement to avoid any confusion for a first day user?

cc @N-Usha

ericsampson:
Here's a proposal that might address all those questions, and is still simple for the new user; these 4 individual inputs would be for more advanced users.

- existing `creds` input works as it does now (except that it's optional)
[...]

PR #108, opened 2021-05-04
[FEATURE] Ability to Pass Authentication Parameters Individually

anatbal:
**As mentioned in #39:**
Allow login with environment variables instead of json metadata file: passing in the Subscription ID, Tenant ID, Client ID, and Client Secret as individual parameters instead of bundled in a single JSON object (`creds` today). 
This would allow for easier use in pipelines that use Terraform (as it requires them split out). It would also make secret rotation easier as you don't have to rebuilt the JSON object.
 ```yaml
 - uses: azure/login@v1.1
       with:
         creds_client_id: ${{ secrets.CLIENT_ID }}
         creds_client_secret: ${{ secrets.CLIENT_SECRET }}
         creds_subscription_id: ${{ secrets.SUBSCRIPTION_ID }}
         creds_tenant_id: ${{ secrets.TENANTID }}
 ```


src/main.ts
```diff
@@ -39,13 +39,13 @@ async function main() {
...
-        let resourceManagerEndpointUrl = secrets.getSecret("$.resourceManagerEndpointUrl", false);
+        let creds = core.getInput('creds', { required: false });
+        let secrets = creds ? new SecretParser(creds, FormatType.JSON) : null;
+        let servicePrincipalId = core.getInput('creds_client_id', { required: false }) ? core.getInput('creds_client_id', { required: false}) : secrets.getSecret("$.clientId", false);
+        let servicePrincipalKey = core.getInput('creds_client_secret', { required: false }) ? core.getInput('creds_client_secret', { required: false}) : secrets.getSecret("$.clientSecret", false);
+        let tenantId = core.getInput('creds_tenant_id', { required: false }) ? core.getInput('creds_tenant_id', { required: false }) : secrets.getSecret("$.tenantId", false);
+        let subscriptionId = core.getInput('creds_subscription_id', { required: false }) ? core.getInput('creds_subscription_id', { required: false }) : secrets.getSecret("$.subscriptionId", false);
+        let resourceManagerEndpointUrl = creds ? secrets.getSecret("$.resourceManagerEndpointUrl", false) : "https://management.azure.com/";
```
BALAGA-GAYATRI, line 48: @anatbal resourceManagerEndpointUrl is different for sovereign clouds for example us gov-  "resourceManagerEndpointUrl": "https://management.usgovcloudapi.net/", we can't default it to "https://management.azure.com/". Ideally this should also be passed separately in that case.  @N-Usha ,@kmkumaran any workaround for this other than explicitly passing resourceManagerEndpoint.

anatbal: @BALAGA-GAYATRI, I changed the resourceManagerEndpointUrl to also be passed separately and fixed the default value. Let me know if that looks okay to you.

action.yml
```diff
@@ -4,7 +4,19 @@ description: 'Authenticate to Azure and run your Az CLI or Az
 inputs:
```
BALAGA-GAYATRI, line 4: @anatbal creds_resource_manager_endpoint_url should also be added to input fields

anatbal: Thanks @BALAGA-GAYATRI , I fixed it

ericsampson:
bump for re-review @BALAGA-GAYATRI ? :)

kaverma:
@BALAGA-GAYATRI , is this ready to be merged?

BALAGA-GAYATRI:
Hey @anatbal ! Thank you for the contribution towards this action. We have added OIDC login mechanism to the action which now accepts individual parameters as inputs. Please refer [sample workflow](https://github.com/Azure/login#sample-workflow-that-uses-azure-login-action-using-oidc-to-run-az-cli-linux) and [configure OIDC](https://github.com/Azure/login#configure-a-service-principal-with-a-federated-credential-to-use-oidc-based-authentication) for more details. So closing this PR.

PR #120, opened 2021-06-16
Create auto-triage-issues

ishitachawla:
This action automatically labels issues as bug, enhancement or doc using NLP. This action is triggered whenever an issue is created and it only labels unlabeled issues.
DELTA triggers labels issues up to DELTA days back. Its primary purpose is to label pre-existing issues on initial setup. 
CONFIDENCE  skips prediction when below confidence threshold. Any value of confidence can be picked from the interval [-10, 4]. Default value is 2. Note that -10 (or an extremely low value) essentially forces the model to make a prediction for all issues.

### DIFF
--- a/.github/workflows/auto-triage-issues
+++ b/.github/workflows/auto-triage-issues
@@ -1,3 +1,4 @@
+name: "Auto-Labelling Issues"
 on:
   issues:
     types: [opened, edited]
@@ -5,14 +6,14 @@ on:
 jobs:
   auto_label:
     runs-on: ubuntu-latest
-    name: Automatic Github Issue Labeller
+    name: Auto-Labelling Issues
     steps:
     - name: Label Step
       uses: larrylawl/Auto-Github-Issue-Labeller@main
       with:
         GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
         REPOSITORY: ${{github.repository}}
-        DELTA: "1"
+        DELTA: "7"
         CONFIDENCE: "2"
         FEATURE: "enhancement"
         BUG: "bug"
--- a/.github/workflows/auto-triage-issues
+++ b/.github/workflows/auto-triage-issues
@@ -9,7 +9,7 @@ jobs:
     name: Auto-Labelling Issues
     steps:
     - name: Label Step
-      uses: larrylawl/Auto-Github-Issue-Labeller@main
+      uses: larrylawl/Auto-Github-Issue-Labeller@v1.0
       with:
         GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
         REPOSITORY: ${{github.repository}}
@@ -18,4 +18,4 @@ jobs:
         FEATURE: "enhancement"
         BUG: "bug"
         DOCS: "documentation"
-        VERSION: "v1.0"
+        
--- a/.github/workflows/auto-triage-issues
+++ b/.github/workflows/auto-triage-issues
@@ -1,0 +1,20 @@
+on:
+  issues:
+    types: [opened, edited]
+
+jobs:
+  auto_label:
+    runs-on: ubuntu-latest
+    name: Automatic Github Issue Labeller
+    steps:
+    - name: Label Step
+      uses: larrylawl/Auto-Github-Issue-Labeller@main
+      with:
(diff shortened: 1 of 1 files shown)


kanika1894 (commented):
Please add a description to the PR, of what we are trying to do.
If possible, add comments in the Workflow file to explain what is the significance of delta and confidence value.
Also, any reason for keeping it 1 and 2?

.github/workflows/auto-triage-issues
```diff
@@ -0,0 +1,20 @@
+on:
```
kanika1894, line 1: It should have a name field as well

.github/workflows/auto-triage-issues
```diff
@@ -0,0 +1,20 @@
...
+  auto_label:
+    runs-on: ubuntu-latest
+    name: Automatic Github Issue Labeller
+    steps:
+    - name: Label Step
+      uses: larrylawl/Auto-Github-Issue-Labeller@main
+      with:
+        GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+        REPOSITORY: ${{github.repository}}
+        DELTA: "1"
```
kanika1894, line 15: This means it will label just 1 day old issues right?
Let's keep it to 7 days

.github/workflows/auto-triage-issues
```diff
@@ -0,0 +1,20 @@
...
+  issues:
+    types: [opened, edited]
+
+jobs:
+  auto_label:
+    runs-on: ubuntu-latest
+    name: Automatic Github Issue Labeller
+    steps:
+    - name: Label Step
+      uses: larrylawl/Auto-Github-Issue-Labeller@main
```
kanika1894, line 11: Why are we using their @main branch? Don't they have a release version?

.github/workflows/auto-triage-issues
```diff
@@ -0,0 +1,20 @@
...
+      with:
+        GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+        REPOSITORY: ${{github.repository}}
+        DELTA: "1"
+        CONFIDENCE: "2"
+        FEATURE: "enhancement"
+        BUG: "bug"
+        DOCS: "documentation"
+        VERSION: "v1.0"
```
kanika1894, line 20: What's the significance of this version field?

.github/workflows/auto-triage-issues
```diff
@@ -0,0 +1,20 @@
+on:
+  issues:
+    types: [opened, edited]
+
+jobs:
+  auto_label:
+    runs-on: ubuntu-latest
+    name: Automatic Github Issue Labeller
+    steps:
+    - name: Label Step
```
kanika1894, line 10: It can be renamed to something which is more intuitive, such as "Auto labeling issues"

.github/workflows/auto-triage-issues
ishitachawla: Two version of the action are available. This one gets chosen by default

.github/workflows/auto-triage-issues
ishitachawla: Done

.github/workflows/auto-triage-issues
ishitachawla: They dont

kanika1894: They have v1
https://github.com/larrylawl/auto-github-issue-labeller-action/tags

.github/workflows/auto-triage-issues
kanika1894: So it's not something that we should specify as a parameter right?

ishitachawla: Yeah it's optional. We can leave it if we want

ishitachawla: Removed it.

.github/workflows/auto-triage-issues
ishitachawla: Updated it.

kanika1894 (approved):

hubjacek:
> [CLA assistant check](https://cla.opensource.microsoft.com/Azure/login?pullRequest=120) All CLA requirements met.

PR #132, opened 2021-08-25
Update packages to mitigate various CVEs

ChaosCypher:
The package underscore from 1.13.0-0 and before 1.13.0-2, from 1.3.2 and before 1.12.1 are vulnerable to Arbitrary Code Injection via the template function, particularly when a variable property is passed as an argument as it is not sanitized.

Some of the dependencies required a manual audit, once `xmldom 0.7.0` is released, this will resolve the remaining CVEs.

EDIT: tagging CODEOWNERS @kaverma @kanika1894 @BALAGA-GAYATRI @pulkitaggarwl

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -707,6 +707,13 @@
         "jsonpath": "^1.0.2",
         "xmldom": "^0.1.27",
         "xpath": "0.0.27"
+      },
+      "dependencies": {
+        "xmldom": {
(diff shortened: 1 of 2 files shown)


BALAGA-GAYATRI:
@ChaosCypher. I'm trying to understand why we need extra packages `underscore` and  `xmldom` as the actions is working fine without them also 🤔 💭 .

PR #140, opened 2021-09-13
Update Readme to improve examples for hand crafting credentials

IronSean:
The clientSecret is not a GUID, and this being labelled as such caused confusion when I tried to manually add a client secret and was met with a SecretId which was a GUID and a SecretValue which was not a GUID. Using the SecretId GUID from the Azure UI would not work.

Also, resourceManagerEndpointUrl was required but not shown in any examples. 

Finally, the value for it was hard to determine, as using several azure cli commands did not display it.


kanika1894 (approved):

kanika1894:
Thanks for your contribution @IronSean

PR #159, opened 2021-10-25
Adding federated token logs & optional audience parameter

BALAGA-GAYATRI:
Added federated token logs - sample output:
Screenshot 2021-10-25 at 7 00 15 PM

Also, added audience parameter (which will be optional) and defaults to ‘api://AzureADTokenExchange’


README.md
```diff
@@ -121,7 +121,7 @@ jobs:
           az group list
           pwd 
 ```
-This action supports login az powershell as well for both windows and linux runners by setting an input parameter `enable-AzPSSession: true`. Below is the sample workflow for the same using the windows runner. Please note that powershell login is not supported in Macos runners.
+Users can also specify `audience` field for access-token in the input parameters of the action. If not specified, it is defaulted to `api://AzureADTokenExchange`. This action supports login az powershell as well for both windows and linux runners by setting an input parameter `enable-AzPSSession: true`. Below is the sample workflow for the same using the windows runner. Please note that powershell login is not supported in Macos runners.
```
kaverma, line 124: nit: we should have "This action supports login... " in a new line

kaverma (approved):

kaverma:
We need to update the documentation/readme to include the audience field

kaverma (changes requested):
@BALAGA-GAYATRI, @N-Usha , should we have a security review to ensure whether we are good with all these claims in logs that anyone can see?

BALAGA-GAYATRI:
Printing token logs changes are removed from this PR and will be done post security approval.

PR #171, opened 2021-11-11
Adding logs to surface AZ-CLI and powershell errors

BALAGA-GAYATRI:
Screenshot 2021-11-11 at 5 50 02 PM


lib/main.js
```diff
@@ -180,12 +208,11 @@ function main() {
         }
         catch (error) {
             if (!isAzCLISuccess) {
-                core.error("Az CLI Login failed. Please check the credentials. For more information refer https://aka.ms/create-secrets-for-GitHub-workflows");
+                core.setFailed("Az CLI Login failed. Please check the credentials. For more information refer https://aka.ms/create-secrets-for-GitHub-workflows");
```
kaverma, line 211: what's the difference between .error and .setFailed

src/main.ts
```diff
@@ -10,6 +11,23 @@ var azPSHostEnv = !!process.env.AZUREPS_HOST_ENVIRONMENT ? `${process.env.AZUREP
...
     try {
+        //Options for error handling
+        let commandStdErr = false;
+        const loginOptions: ExecOptions = {
+            silent: true,
+            ignoreReturnCode: true,
+            failOnStdErr: true,
+            listeners: {
+                stderr: (data: Buffer) => {
+                    let error = data.toString().replace('ERROR','');
```
kaverma, line 22: we want to replace only the starting ERROR substring and not all ERROR mentions right?

src/main.ts
```diff
@@ -10,6 +11,23 @@ var azPSHostEnv = !!process.env.AZUREPS_HOST_ENVIRONMENT ? `${process.env.AZUREP
...
+                    if(error && error.trim().length !== 0)
```
kaverma, line 23: nit: formatting - space after if

kaverma (approved):

lib/main.js
BALAGA-GAYATRI: execution stops here if `setFailed` otherwise it continues.

src/main.ts
BALAGA-GAYATRI: done

src/main.ts
BALAGA-GAYATRI: Made small change. It replaces only first occurrence in the beginning of the message  now.

PR #183, opened 2021-12-01
Allow OIDC login via `creds`

colindembovsky:
Currently, if you you want to use OIDC, you have to pass in 3 parameters as follows:

```yml
  steps:
    - name: OIDC Login to Azure Public Cloud with AzPowershell (enableAzPSSession true)
      uses: azure/login@v1
      with:
        subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }} 
```

This PR allows you to use the `creds` parameter. Instead of configuring login to OIDC or credentials based on the presence of the `creds` property, the logic now checks if the `creds` JSON contains a `clientSecret` property. If this is present, credentials are used. If not, then OIDC is used.

This means you can now log in using OIDC as follows:

```yml
  steps:
    - name: OIDC Login to Azure Public Cloud with AzPowershell (enableAzPSSession true)
      uses: azure/login@v1
      with:
```

where the secret `AZURE_CREDENTIALS` is specified as follows:

```json
{ 
  "clientId": "<clientId>",
  "subscriptionId": "<sbuscriptionId>",
  "tenantId": "<tenantId>"
}
```


BALAGA-GAYATRI:
Hey @colindembovsky. Thanks for contributing to this action. :) In fact, initially we thought of the same design but here are a few implications- 
1. Encouraging OIDC more is the primary goal for us in the future.
2. There has been more requirement for the provision to use individual creds and secrets instead of json. https://github.com/Azure/login/issues/39
3. More importantly,[ Using a structured object as a secret](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-secrets) is not recommended in our security guidelines. So our attempt is towards converting users more towards OIDC and individual creds.

So to fit in the above requirements and provide the existing workflows to run smoothly, the current design was chosen.

Issue #192, opened 2022-01-04
Azure login action does not function if the client secret starts with hyphen


chapitos:
If the Azure creds clientSecret starts with hyphen the Azure login action will fail with the following error:

`Error: : argument --password/-p: expected one argument`

It seems that the clientSecret is passed to az login using -p and not using -p= which is needed according to az login documentation if the clientSecret starts with a hyphen.

AjayKaleMS:
Hi @chapitos  - We have implemented a fix in a forked branch. We would like you to try your workflow on below action instead of original one(azure/login@v1)

`AjayKaleMS/login@releases/v1
`
If this action works for you for secrets with and without hyphen start, then we can make the changes in original Azure/Login.

Thanks in advance for your co-operation.

chapitos:
@AjayKaleMS: have you tested your code fix?

AjayKaleMS:
Yes @chapitos - We have already tested it but still we want to make it sure if it is indeed working for the inputs you are facing issue with.
Please do let us know result from your side.
Thanks.

chapitos:
@AjayKaleMS your fix is not working so I am not sure how it passed your tests. The action is throwing now a different error: AADSTS7000215: Invalid client secret is provided. 

Your change is putting the secret between quotes, which is not correct (quotes are perceived as part of the secret). Please see my comment on your PR. According to az login documentation, passwords starting with '-' need to be given with -p=secret and not -p secret. This method will also work for other secrets. So in essence, you do not need if-else logic, just use -p=secret when building arguments

AjayKaleMS:
Thanks @chapitos for quick response on this. Actually we manually inserted '-' at the beginning of our correct client secret and this removed the original error `'Error: : argument --password/-p: expected one argument'` . This was the reason we wanted it to be tested with your inputs.

As per your comments, I have made the changes which is secret is passed in the `-p=<secret>`

Request you please test the changes again on the same below branch and kindly let us know it it works.

`AjayKaleMS/login@releases/v1
`
Thanks again.

AjayKaleMS:
@chapitos - could you please try again as there was a block completion error - I corrected it now.

Below is final code block to handle secrets:

```
if (enableOIDC) {
                commonArgs = commonArgs.concat("--federated-token", federatedToken);
            }
            else {
                  commonArgs = commonArgs.concat("-p=", servicePrincipalKey);
            }   
```

chapitos:
[...]

PR #196, opened 2022-01-12
Fix for issue-192 - Azure login action does not function if the client secret starts with hyphen

AjayKaleMS:
Fix for Azure login action does not function if the client secret starts with hyphen

https://github.com/Azure/login/issues/192

Handled the code separately for secret with hyphen and without hyphen.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -189,8 +189,15 @@ function main() {
             if (enableOIDC) {
                 commonArgs = commonArgs.concat("--federated-token", federatedToken);
             }
-            else {
-                  commonArgs = commonArgs.concat("-p=", "\"" + servicePrincipalKey + "\"");
+           else {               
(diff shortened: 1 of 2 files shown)


lib/main.js
```diff
@@ -189,8 +189,16 @@ function main() {
             if (enableOIDC) {
                 commonArgs = commonArgs.concat("--federated-token", federatedToken);
             }
-            else {
-                commonArgs = commonArgs.concat("-p", servicePrincipalKey);
+            else {               
+                    if(servicePrincipalKey.startsWith("-"))
+                    {
+                        commonArgs = commonArgs.concat("-p", "\"" + servicePrincipalKey + "\"");
```
BALAGA-GAYATRI, line 195: Why are we having the separate condition for with and without hyphen(-). Can't we make it generic? @AjayKaleMS

PR #208, opened 2022-03-04
Update ClientSecret in the example to not call itself a GUID

IronSean:
If manually crafting or verifying a Service Principal credential JSON object, it is helpful to know that only the `clientId` is formatted as a GUID, and the `clientSecret` is just an alphanumeric string.


BALAGA-GAYATRI:
Hey @IronSean , I'm trying to understand better here. You are saying making it `<CLIENT-SECRET>` instead of `<GUID>` will help the users understand the format which is used for processing?

PR #221, opened 2022-05-02
Handling warnings and adding OIDC promotion message

### LINKED ISSUE #216: Error `Unable to encode the output with cp1252 encoding.` is shown when logging in to a subscription which name contains Japanese characters and on Windows runner
When trying to log in with a subscription which name contains Japanese characters by the following simple workflow, the logging in is failed with the encoding error like the following screenshot.

Workflow:
```yaml
name: Azure Login Test

on:
  workflow_dispatch:

jobs:
  login:
    runs-on: windows-latest # using Windows runner
    steps:
    - uses: azure/login@v1
      with:
```

Error screenshot:
image

BALAGA-GAYATRI:
Resolves #216 
- Warnings will not be thrown as errors
- Adding an additional OIDC promotion message when a user opts for Non-OIDC login
Screenshot 2022-05-02 at 12 46 27 PM

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -45,11 +45,11 @@ function main() {
                 listeners: {
                     stderr: (data) => {
                         let error = data.toString();
-                        let isWarning = error.toLowerCase().startsWith('warning');
-                        let isError = error.toLowerCase().startsWith('error');
+                        let startsWithWarning = error.toLowerCase().startsWith('warning');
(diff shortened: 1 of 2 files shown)


tauhid621 (approved):
Looks good. Left some comments at release [PR](https://github.com/Azure/login/pull/222#pullrequestreview-958942969)

tauhid621 (approved):

PR #222, opened 2022-05-02
Handling warnings and adding OIDC promotion message release PR

### LINKED ISSUE #216: Error `Unable to encode the output with cp1252 encoding.` is shown when logging in to a subscription which name contains Japanese characters and on Windows runner

BALAGA-GAYATRI:
Resolves #216 
- Warnings will not be thrown as errors
- Adding an additional OIDC promotion message when a user opts for Non-OIDC login
 Screenshot 2022-05-02 at 12 46 27 PM

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -17,8 +17,8 @@ async function main() {
             listeners: {
                 stderr: (data: Buffer) => {
                     let error = data.toString();
-                    let isWarning = error.toLowerCase().startsWith('warning')
-                    let isError = error.toLowerCase().startsWith('error')
+                    let isWarning = error.toLowerCase().startsWith('warning');
+                    let isError = error.toLowerCase().startsWith('error');
                     // printing ERROR
                     if (error && error.trim().length !== 0 && !isWarning) {
                         if(isError) {
--- a/src/main.ts
+++ b/src/main.ts
@@ -15,7 +15,7 @@ async function main() {
         // let commandStdErr = false;
         const loginOptions: ExecOptions = {
             silent: true,
-            ignoreReturnCode: true,
+            // ignoreReturnCode: true,
             // failOnStdErr: true,
             listeners: {
                 stderr: (data: Buffer) => {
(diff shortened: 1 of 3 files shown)


src/main.ts
```diff
@@ -12,22 +12,20 @@ var azPSHostEnv = !!process.env.AZUREPS_HOST_ENVIRONMENT ? `${process.env.AZUREP
...
-                    //removing the keyword 'ERROR' to avoid duplicates while throwing error
-                    if (error.toLowerCase().startsWith('error')) {
-                        error = error.slice(5);
-                    }
-                    // printing error
-                    if (error && error.trim().length !== 0) {
-                        commandStdErr = true;
-                        core.error(error);
+                    let isWarning = error.toLowerCase().startsWith('warning');
+                    let isError = error.toLowerCase().startsWith('error');
```
tauhid621, line 21: NIT : Can be renamed with `startsWithError`.

src/main.ts
```diff
@@ -12,22 +12,20 @@ var azPSHostEnv = !!process.env.AZUREPS_HOST_ENVIRONMENT ? `${process.env.AZUREP
...
+                    // printing ERROR
+                    if (error && error.trim().length !== 0 && !isWarning) {
+                        if(isError) {
+                            //removing the keyword 'ERROR' to avoid duplicates while throwing error
+                            error = error.slice(5);
+                        }
+                        core.setFailed(error);
```
tauhid621, line 28: Earlier it was `core.error(error)`. What is the difference?

tauhid621 (approved):
Looks Good. Left NIT comments.

src/main.ts
BALAGA-GAYATRI: The latter is used to set the exit codes and mark them as failed. In our case both are the same. That shouldn't be a problem.

src/main.ts
BALAGA-GAYATRI: Done 👍

PR #224, opened 2022-05-24
Allow obtaining subscription id both from 'subscription-id' as well as from 'creds'

tomas-pajurek:
Currently, it is not possible use the action in this way:

```yml
- uses: azure/login@v1
  with:
    subscription-id: ${{ env.AZURE_SUBSCRIPTION_ID }}
```
where `secrets.AZURE_CREDENTIALS` = 

```json
{
    "tenantId": "f6749845-e0f4-4d98-8ea4-8910b615cd0e",
    "clientId": "c396b6d6-bf41-4ee1-81c9-0d7707db51f1",
    "clientSecret": "..."
}
```

Expected behavior: subscription id is taken from `subscription-id` input. 
Actual behavior: login fails because no subscription id is set.

This PR fixes this issue by only taking subscription id from `creds` if `subscription-id` is not set.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -74,8 +74,8 @@ async function main() {
         var enableOIDC = true;
         var federatedToken = null;
 
-        // If any of the individual credentials (clent_id, tenat_id, subscription_id) is present.
-        if (servicePrincipalId || tenantId || subscriptionId) {
+        // If any of the individual credentials (clent_id, tenat_id) is present.
+        if (servicePrincipalId || tenantId) {
 
             //If few of the individual credentials (clent_id, tenat_id, subscription_id) are missing in action inputs.
             if (!(servicePrincipalId && tenantId && (subscriptionId || allowNoSubscriptionsLogin)))
@@ -88,7 +88,11 @@ async function main() {
                 servicePrincipalId = secrets.getSecret("$.clientId", true);
                 servicePrincipalKey = secrets.getSecret("$.clientSecret", true);
                 tenantId = secrets.getSecret("$.tenantId", true);
-                subscriptionId = secrets.getSecret("$.subscriptionId", true);
+
+                if(!subscriptionId) {
+                    subscriptionId = secrets.getSecret("$.subscriptionId", true);
+                }
+
                 resourceManagerEndpointUrl = secrets.getSecret("$.resourceManagerEndpointUrl", false);
             }
             else {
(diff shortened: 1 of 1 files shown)


BALAGA-GAYATRI:
Hello @tomas-pajurek 
We had that design to distinguish OIDC (using creds object) and Non OIDC(using individual parameters). Also if you use the creds object generated using [this](https://github.com/Azure/login#configure-a-service-principal-with-a-secret), it guarantees `subscription-id` in the object. You need not care about passing explicitly. If you want to pass individual parameters then pass 3 of them and it will login you with OIDC mechanism.

tomas-pajurek:
Hello Balaga,

thanks for the response.

Can you please explain rationale behind this design? What is the purpose of distinguishing between OIDC and non-OIDC cases in such a strict way?

Frequently, there is a scenario where the service principal is needed to manage resources in multiple subscriptions. In such scenarios, it would be nice to have the credentials stored a Git Hub Actions Secret (without subscription ID) and supply subscription ID in another way. 

In general, I do not see any benefit of forcing users to specify service principal's identity (client id, client secret, tenant id) and subscription id which is actually not related to that identity together.

BALAGA-GAYATRI:
>Can you please explain rationale behind this design? What is the purpose of distinguishing between OIDC and non-OIDC cases in such a strict way? 

We are moving forward with having OIDC as default and promoting it along with using individual parameters.

> In general, I do not see any benefit of forcing users to specify service principal's identity (client id, client secret, tenant id) and subscription id which is actually not related to that identity together.

In a way there is. We create an App in AD and give contributor access from the subscription we chose. [Readme](https://github.com/Azure/login#configure-a-service-principal-with-a-federated-credential-to-use-oidc-based-authentication).

tomas-pajurek:
Thanks again for the response.

However, I mostly do not understand the provided reasoning. Having OIDC as a recommended and somehow default option definitely makes sense. But it is not in conflict with changes I am proposing. Other thing I do not understand is the motivation to force users to use service principals in specific way. I understand that it is desired to guide new users in the best (OIDC) direction. However, given that the service principals exists much longer than this GitHub Action and/or OIDC option, there are many users already using service principals in their CI pipelines in different ways. Compared to this, I find the argument that the README of this repo recommends some specific way how to create an SP to be weak. To sum up, I think that this design is makes adoption of this GitHub Action by existing CI pipelines less ergonomic while not bringing significant benefits.

Please, take this as a constructive feedback. If you are sure about the current design, there is no need to discuss this further and feel free to close this PR.

BALAGA-GAYATRI:
@tomas currently we can't go ahead with this implementation due to design limitations. So closing it for now. Will definitely reopen in case of any re-considerations. Thank you for the feedback.

tomas:
@BALAGA-GAYATRI  Please don't mention me again. I have nothing to do with this.

PR #245, opened 2022-09-09
[BUG FIX] Version pattern quickfix releases

BALAGA-GAYATRI:
The [version_pattern_regex](https://github.com/Azure/login/blob/master/src/PowerShell/Constants.ts#L4) in the action, it was accepting only single-digit major, minor, and patch version numbers. Since the new release was 2.10, the validation is failing and failing the action. So fixing that regex is working now.


Sindhuja365 (approved):

Toppole6 (commented):
@[](https://familycenter.messenger.com/invite/p/WAiL1DQfTfEhWryu6S2b8hcBunEJE/)

PR #247, opened 2022-09-21
Update action.yml description to include the "OIDC" keyword

N-Usha:


BALAGA-GAYATRI:
What is the reason for the change? Is the below screenshot the only place where this change gets reflected? Also why not include SPN along with OIDC ?
Screenshot 2022-09-22 at 12 38 16 AM

N-Usha:
Want the action to appear when they "keyword" OIDC is used for searching across this above window & also marketplace. AFAIK we look for keywords in Action names & descriptions alone.

And authentication with OIDC (OIDC+SPN, OIDC+MSI) is eventually what we want to promote. Auth w/o OIDC is the other alternative that we support but IMHO we don't need to call that out or promote expicitly.

BALAGA-GAYATRI (approved):

PR #272, opened 2022-11-16
Improved debug output

bb-froggy:
The debug output is wrong, seems to be a copy & paste error. There is another debug output later on that also says 'az cli version used', so it is easy to confuse the two.

The change to line 239 seems to be some GitHub whitespace thing, I didn't intend that one, but it looks like it doesn't hurt.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -67,7 +67,7 @@ function main() {
             core.exportVariable('AZURE_HTTP_USER_AGENT', userAgentString);
             core.exportVariable('AZUREPS_HOST_ENVIRONMENT', azurePSHostEnv);
             azPath = yield io.which("az", true);
-            core.debug(`az cli version used: ${azPath}`);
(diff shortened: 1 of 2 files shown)


BALAGA-GAYATRI:
Hi, Can you please make the changes to the .js counterpart of the above file? Thanks in advance

bb-froggy:
Done and I also rebased to the newest master

BALAGA-GAYATRI:
Amazing!! Thanks for the quick response.

BALAGA-GAYATRI (approved):

bb-froggy:
GitHub says that `releases/v1` and `master` [are entirely different commit histories](https://github.com/Azure/login/compare/master...releases/v1) and thus does not let me create a Pull Request. I did not investigate why it thinks that, as I don't know about this repository's structure.

PR #313, opened 2023-05-05
Close the issue with label `needs-author-feedback` for 20 days without activity

MoChilia:
Automatically check whether the issue has been labeled `needs-author-feedback` for 20 days without activity. If yes, then close the issue with a comment. If the author has updated the issue, then remove the `needs-author-feedback` label.

Set `days-before-stale` as `-1`, so that it will not reset any other stale label.
Set `needs-author-feedback` as `stale-issue-label`, so that if the issue has any update, it will remove the `needs-author-feedback` label. This ensures that only issues without any update will be closed. 
Set `only-issue-labels` as `needs-author-feedback`, so that only issues with `needs-author-feedback` label are searched.

### DIFF
--- a/.github/workflows/defaultLabels.yml
+++ b/.github/workflows/defaultLabels.yml
@@ -34,3 +34,14 @@ jobs:
           days-before-stale: 14
           days-before-close: -1
           operations-per-run: 100
+
+      - uses: actions/stale@v3
+        name: Closing issue with no feedback for 7 days
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


YanaXu:
- We'll label `needs-author-feedback` manually.
- If an issue has been labeled with `needs-author-feedback` for 20 days with no activity, it'll be closed.
- If an issue has been labeled with `needs-author-feedback` within 20 days and the author provides a feedback, the label `needs-author-feedback` will be removed automatically.

YanaXu (approved):

PR #319, opened 2023-05-18
Support Managed Identity login for self-hosted runner

MoChilia:
### Description
This PR is going to support both system- and user- assigned managed identity login for self-hosted runners on Azure VM. 

We add two new input options:
`enable-managed-identity`: 'Set this value to true to enable Managed Identity for your action. Requires a self-hosted Github runner on an Azure VM with Managed Identity enabled.'
`user-managed-identity-resource-id`: 'Set this value to the Resource ID of your user managed identity. If using a system managed identity, leave this empty. This setting has no effect if `enable-managed-identity` is false.'

CLI Part is ready now.

### Test workflows:
[System-assigned managed identity test](https://github.com/MoChilia/ActionsDemo/actions/workflows/TestSystemManagedIdentity.yml)
[User-assigned managed identity test](https://github.com/MoChilia/ActionsDemo/actions/workflows/TestUserManagedIdentity.yml)

Related to issue #56 
With reference to pr #68

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,6 @@ Note:
 ## Sample workflow that uses Azure login action to run az cli
 
 ```yaml
-
 # File: .github/workflows/workflow.yml
 
 on: [push]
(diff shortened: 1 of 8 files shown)


htekdev:
This would be amazing!! Directly completes with Azure DevOps Self-Hosted Managed Identity Authentication

PR #321, opened 2023-05-24
Support OIDC for sovereign clouds

MoChilia:
### Description
This pr is going to support OIDC based authentication for sovereign clouds.

- Fix issue https://github.com/Azure/login/issues/298
- Close issue https://github.com/Azure/login/issues/257
- Close pr https://github.com/Azure/login/pull/258

### Test workflows

- [Positive Test OIDC for Azure China Cloud](https://github.com/MoChilia/ActionsDemo/actions/runs/5066636230/jobs/9096689312)
- [Positive Test OIDC for Azure Cloud](https://github.com/MoChilia/ActionsDemo/actions/runs/5066051305/jobs/9095354865)
- [Negative Test OIDC for Azure China Cloud: Write permission not given](https://github.com/MoChilia/ActionsDemo/actions/runs/5066276391/jobs/9095874627)
- [Negative Test OIDC for Azure China Cloud: No federated credentials is configured](https://github.com/MoChilia/ActionsDemo/actions/runs/5066500495/jobs/9096378581)

PR #328, opened 2023-05-29
Use `creds` as supplementary of individual parameters

MoChilia:
### Description
This pr is going to read `creds` from Json, the items in `creds` will not overwrite the individual parameters `client-id`, `tenant-id` and `subscription-id`, but as supplementary. 

- In the previous code, `creds` is not compatible with individual parameters. We use `creds` for login using service principal with secret and individual parameters for `OIDC` login:
https://github.com/Azure/login/blob/990b22f2bceb30f6bb94f1713b46a0be2c707d93/src/main.ts#L78-L93
In the new version, we aim to fetch the user's input credentials as much as possible:
https://github.com/Azure/login/blob/b3bc91bd1727e02f5e0c619bde1ede0aa372a02a/src/common/LoginConfig.ts#L35-L42

- We remove the boolean `enableOIDC` and use `servicePrincipalKey` to decide whether OIDC is enabled.

- We also remove the adjudication for undefined `federatedToken` in L122, since the error will be caught in L118. (Related to https://github.com/Azure/login/pull/321)
https://github.com/Azure/login/blob/990b22f2bceb30f6bb94f1713b46a0be2c707d93/src/main.ts#L115-L129

### DIFF
--- a/__tests__/PowerShell/ServicePrinicipalLogin.test.ts
+++ b/__tests__/PowerShell/ServicePrinicipalLogin.test.ts
@@ -1,11 +1,17 @@
 import { ServicePrincipalLogin } from '../../src/PowerShell/ServicePrincipalLogin';
+import { LoginConfig } from '../../src/common/LoginConfig';
 
 jest.mock('../../src/PowerShell/Utilities/Utils');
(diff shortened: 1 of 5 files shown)


MoChilia:
Closed this pr, it is included in https://github.com/Azure/login/pull/336.

PR #331, opened 2023-06-01
docs: fix type in creds attribute

alegrey91:
This PR fixes the secrets names in `creds` attribute.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,7 +231,8 @@ The command should output a JSON object similar to this:
 ```yaml
   - uses: Azure/login@v1
     with:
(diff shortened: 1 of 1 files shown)


YanaXu:
Hi @alegrey91 ,  the line you want to update is just an example. Could you provide a reason why you think "AZURE_AD_CLIENT_ID " is better than "CLIENT_ID"? Thanks.

alegrey91:
> Hi @alegrey91 , the line you want to update is just an example. Could you provide a reason why you think "AZURE_AD_CLIENT_ID " is better than "CLIENT_ID"? Thanks.

Hi! Most of the examples are using `AZURE_AD_CLIENT_ID` instead of `CLIENT_ID`.
This can create confusion IMO.

YanaXu:
The suggestion is included in the latest release. Thanks for your advice.

PR #336, opened 2023-06-08
Support Managed Identity login for self-hosted runner

MoChilia:
### Description
This PR is going to support both system- and user- assigned managed identity login for self-hosted runners on Azure VM. 

#### What's new
- The Action provides a parameter `auth-type` with value list `[SERVICE_PRINCIPAL, IDENTITY]` to identify the type of authentication.
  1. If `auth-type: SERVICE_PRINCIPAL` with `clientId`, `tenantId` and `clientSecret` detected in your input, we will attempt to login by using service principal with the secret.
  ```yaml
  #login with secret
  - uses: azure/login@v1
        with:
          auth-type: SERVICE_PRINCIPAL
  ```
  2. If `auth-type: SERVICE_PRINCIPAL` with `clientId` and `tenantId` detected in your input, we will attempt to login by using OIDC.
  ```yaml
  #login with OIDC
  - uses: azure/login@v1
      with:
          subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
          auth-type: SERVICE_PRINCIPAL
  ```
  3. If `auth-type: IDENTITY` with `clientId` detected in your input, we will attempt to login by using user-assigned managed identity.
   ```yaml
   #login with user-assigned managed identity
  - uses: azure/login@v1
      with:
          auth-type: IDENTITY
  ```
  4. If `auth-type: IDENTITY` without `clientId` detected in your input, we will attempt to login by using system-assigned managed identity.
   ```yaml
   #login with system-assigned managed identity
  - uses: azure/login@v1
    with:
          auth-type: IDENTITY
  ```
In order not to introduce breaking change, we set the default value of `auth-type` to be `SERVICE_PRINCIPAL`.

- About the parameter `subscription-id`
This parameter used to be mandatory if `allow-no-subscriptions` is not enabled in previous version:
https://github.com/Azure/login/blob/bed977350c1362eceb34cb326960d8c65b9847e9/src/common/LoginConfig.ts#L79-L81 
However, the two parameters are not relevant. `allow-no-subscriptions` is used to login tenant-level account and `subscription-id` is used to specify which subscription to work. Not specifying a subscription should be allowed, then the Action will use the current, active subscription. Hence the limitation is removed in this version with only warning being reported: https://github.com/Azure/login/blob/19aba32e47c97332540e246165041846831ec3b0/src/Cli/AzureCliLogin.ts#L178-L184

- About reading `creds` from Json
The items in `creds` will not overwrite the individual parameters `client-id`, `tenant-id` and `subscription-id`, but as supplementary.  
In the previous code, `creds` is not compatible with individual parameters. We use `creds` for login using service principal with secret and individual parameters for `OIDC` login:
https://github.com/Azure/login/blob/990b22f2bceb30f6bb94f1713b46a0be2c707d93/src/main.ts#L78-L93
In the new version, we aim to fetch the user's input credentials as much as possible:
https://github.com/Azure/login/blob/b3bc91bd1727e02f5e0c619bde1ede0aa372a02a/src/common/LoginConfig.ts#L35-L42

- The changes for README.md is in https://github.com/Azure/login/pull/344.

### DIFF
--- a/.github/workflows/azure-login-negative.yml
+++ b/.github/workflows/azure-login-negative.yml
@@ -146,7 +146,7 @@ jobs:
         tenant-id: ${{ secrets.OIDC_SP2_TENANT_ID }}
         subscription-id: ${{ secrets.OIDC_SP2_SUBSCRIPTION_ID }}
         allow-no-subscriptions: true
-        enable-AzPSSession: true
+        # enable-AzPSSession: true
 
     - name: Run Azure Cli
       id: cli_3
--- a/.github/workflows/azure-login-negative.yml
+++ b/.github/workflows/azure-login-negative.yml
@@ -374,3 +374,56 @@ jobs:
       with:
           script: |
             core.setFailed('Last action should fail but not. Please check it.')
+
+    # Secret of SP1 in creds will be used to sign in SP2
+    - name: Login with both creds and individual parameters
+      id: login_12
+      continue-on-error: true
+      uses: ./
+      with:
+        creds: ${{secrets.SP1}}
+        client-id: ${{ secrets.OIDC_SP2_CLIENT_ID }}
+        tenant-id: ${{ secrets.OIDC_SP2_TENANT_ID }}
+        subscription-id: ${{ secrets.OIDC_SP2_SUBSCRIPTION_ID }}
+        allow-no-subscriptions: true
+        enable-AzPSSession: true
+    
+    - name: Check Last step failed
+      if: steps.login_12.outcome == 'success'
+      uses: actions/github-script@v3
+      with:
+          script: |
+            core.setFailed('Last action should fail but not. Please check it.')
+
+  VMTest:
+    strategy:
+      matrix:
+        os: [self_linux, self_windows]
+    runs-on: ${{ matrix.os }}
+    environment: Automation test
+
+    steps:
+    - name: 'Checking out repo code'
+      uses: actions/checkout@v3.5.2
+
+    - name: Set Node.js 16.x for GitHub Action
+      uses: actions/setup-node@v1
+      with:
+        node-version: 16.x
+
+    - name: 'Validate build'
+      run: |
+        npm install
+        npm run build
+
+    - name: Login with system-assigned managed identity without auth-type
+      id: login_13
+      continue-on-error: true
+      uses: ./
+    
+    - name: Check Last step failed
+      if: steps.login_13.outcome == 'success'
+      uses: actions/github-script@v3
+      with:
+          script: |
+            core.setFailed('Last action should fail but not. Please check it.')
--- a/.github/workflows/azure-login-negative.yml
+++ b/.github/workflows/azure-login-negative.yml
@@ -146,7 +146,7 @@ jobs:
         tenant-id: ${{ secrets.OIDC_SP2_TENANT_ID }}
         subscription-id: ${{ secrets.OIDC_SP2_SUBSCRIPTION_ID }}
         allow-no-subscriptions: true
-        # enable-AzPSSession: true
+        enable-AzPSSession: true
 
     - name: Run Azure Cli
       id: cli_3
--- a/.github/workflows/azure-login-negative.yml
+++ b/.github/workflows/azure-login-negative.yml
@@ -357,3 +357,20 @@ jobs:
       with:
           script: |
             core.setFailed('Last action should fail but not. Please check it.')
+
+    - name: Login with tenant-level account, without allow-no-subscriptions
+      id: login_11
+      continue-on-error: true
+      uses: ./
+      with:
+        client-id: ${{ secrets.OIDC_SP2_CLIENT_ID }}
+        tenant-id: ${{ secrets.OIDC_SP2_TENANT_ID }}
+        subscription-id: ${{ secrets.OIDC_SP2_SUBSCRIPTION_ID }}
+        enable-AzPSSession: true
+
+    - name: Check Last step failed
+      if: steps.login_11.outcome == 'success'
+      uses: actions/github-script@v3
+      with:
+          script: |
+            core.setFailed('Last action should fail but not. Please check it.')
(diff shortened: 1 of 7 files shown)


src/Cli/AzureCliLogin.ts
```diff
@@ -95,11 +81,114 @@ export class AzureCliLogin {
         console.log(`Done registering cloud: "${this.loginConfig.environment}"`)
     }
 
+    async loginWithSecret() {
+        if (!(this.loginConfig.servicePrincipalId && this.loginConfig.tenantId && this.loginConfig.servicePrincipalKey) || this.isSuccess) {
```
YanaXu, line 85: Switch the 2 conditions.
Add a debug log in this "if".

src/Cli/AzureCliLogin.ts
```diff
@@ -95,11 +81,114 @@ export class AzureCliLogin {
...
+            "-p", this.loginConfig.servicePrincipalKey
+        ];
+        if (this.loginConfig.allowNoSubscriptionsLogin) {
+            commonArgs.push("--allow-no-subscriptions");
+        }
+        try {
+            await this.executeAzCliCommand('login', commonArgs, true, this.loginOptions);
+            await this.setSubscription();
+            this.isSuccess = true;
+            console.log('Az cli login succeed by using service principal with secret.');
```
YanaXu, line 101: Extract a new method for login, the start line is "try".
The new method will accept commonArgs and call setSubscription.

src/Cli/AzureCliLogin.ts
```diff
@@ -95,11 +81,114 @@ export class AzureCliLogin {
...
+
+    async loginWithOIDC() {
+        if (!(this.loginConfig.servicePrincipalId && this.loginConfig.tenantId) || this.isSuccess) {
+            return;
+        }
+        console.log('Attempting az cli login by using OIDC...');
+        await this.loginConfig.getFederatedToken();
+        let commonArgs = ["--service-principal",
+            "-u", this.loginConfig.servicePrincipalId,
+            "--tenant", this.loginConfig.tenantId,
```
YanaXu, line 116: Use the same format for the parameters' names, "-" or "--".

.github/workflows/azure-login-negative.yml
```diff
@@ -146,7 +146,7 @@ jobs:
         tenant-id: ${{ secrets.OIDC_SP2_TENANT_ID }}
         subscription-id: ${{ secrets.OIDC_SP2_SUBSCRIPTION_ID }}
         allow-no-subscriptions: true
-        enable-AzPSSession: true
+        # enable-AzPSSession: true
```
YanaXu, line 149: Please update the source code to avoid skipping Azure PowerShell Login.

src/Cli/AzureCliLogin.ts
```diff
@@ -24,39 +31,25 @@ export class AzureCliLogin {
                 }
             }
         };
-        await this.executeAzCliCommand("--version", true, execOptions);
+
+        await this.executeAzCliCommand('--version', [], true, execOptions);
```
jiasli, line 35: `--version` is not an Azure CLI command. We may move it to `args`.

src/Cli/AzureCliLogin.ts
```diff
@@ -24,39 +31,25 @@ export class AzureCliLogin {
...
         };
-        await this.executeAzCliCommand("--version", true, execOptions);
+
+        await this.executeAzCliCommand('--version', [], true, execOptions);
         core.debug(`az cli version used:\n${output}`);
 
         this.setAzurestackEnvIfNecessary();
 
-        await this.executeAzCliCommand(`cloud set -n "${this.loginConfig.environment}"`, false);
+        await this.executeAzCliCommand('cloud set', ['-n', `${this.loginConfig.environment}`], false);
```
jiasli, line 40: Consider move `cloud set` to `args` as well. See https://github.com/Azure/cli/pull/105#discussion_r1226265447

src/Cli/AzureCliLogin.ts
```diff
@@ -7,13 +7,20 @@ import * as io from '@actions/io';
...
     constructor(loginConfig: LoginConfig) {
         this.loginConfig = loginConfig;
+        this.loginOptions = defaultExecOptions();
+        this.isSuccess = false;
     }
 
     async login() {
         this.azPath = await io.which("az", true);
+        if (!this.azPath) {
+            throw new Error("az cli is not found in the runner.");
```
jiasli, line 22: `az cli` is never an official name. It should always be `Azure CLI`.

README.md
```diff
@@ -143,6 +156,44 @@ jobs:
...
+```yaml
+# File: .github/workflows/workflow.yml
+on: [push]
+name: Azure System-assigned Managed Identity sample
+jobs:
+  build-and-deploy:
+    runs-on: self-hosted
+    steps:
+    - name: Azure Login
+      uses: azure/login@v1
```
jiasli, line 170: I am not sure if it is a good idea to implicitly use managed identity. As an alternative, we may add additional `with` arguments indicating `azure/login@v1` should use managed identity, such as `identity: true`. Better to confirm with our PMs.

src/Cli/AzureCliLogin.ts
```diff
@@ -95,11 +88,86 @@ export class AzureCliLogin {
         console.log(`Done registering cloud: "${this.loginConfig.environment}"`)
     }
 
+    async loginWithSecret() {
+        if (this.isSuccess || !(this.loginConfig.servicePrincipalId && this.loginConfig.tenantId && this.loginConfig.servicePrincipalKey)) {
+            core.debug('Skip login with secret.');
+            return;
+        }
```
jiasli, line 95: It is not a programming convention to put the functionality for routing into each execution function `loginWith...`. Execution functions should be definitive and mutually exclusive.

We may take Azure CLI's source code as references:

https://github.com/Azure/azure-cli/blob/d331edf76e5866cda1e6e9641d935256fc4f03ff/src/azure-cli/azure/cli/command_modules/profile/custom.py#L119-L122

https://github.com/Azure/azure-cli/blob/060b414d8ac5e4d8f4fc17cc2e23c2154c762226/src/azure-cli-core/azure/cli/core/_profile.py#L146-L159

bitcoinfinancier (approved):

PR #342, opened 2023-08-02
Update README.md for replacing `--sdk-auth` with `--json-auth`

MoChilia:
### Description
https://github.com/Azure/azure-cli/pull/26572 has added an alias `--json-auth` for `--sdk-auth` in Azure CLI versions >= 2.51.0. Since `--sdk-auth` might be deprecated in near future, we should use `--json-auth` instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,7 +196,7 @@ Refer to the [Azure Stack Hub Login Action Tutorial](https://learn.microsoft.com
 
 For using any credentials like Azure Service Principal, Publish Profile etc add them as [secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets) in the GitHub repository and then use them in the workflow.
 
-Follow the following steps to configure Azure Service Principal with a secret in the scope of `resource-group` as the role of `contributor`:
+Follow the following steps to configure Azure Service Principal with a secret at the scope of `{resource-group}` as the role of `contributor`:
 
 - Define a new secret under your repository settings, Add secret menu
 - Store the output of the below [Azure CLI](https://learn.microsoft.com/cli/azure/?view=azure-cli-latest) command as the value of secret variable, for example 'AZURE_CREDENTIALS'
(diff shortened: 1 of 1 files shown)


YanaXu (approved):

README.md
```diff
@@ -225,7 +225,7 @@ The command should output a JSON object similar to this:
 ```
 
 - Now in the workflow file in your branch: `.github/workflows/workflow.yml` replace the secret in Azure login action with your secret (Refer to the example above)
-- Note: The above `az ad sp create-for-rbac` command will give you the `--sdk-auth` deprecation warning. As we are working with CLI for this deprecation process, we strongly recommend users to use this `--sdk-auth` flag as the result dictionary output changes and not accepted by login action if `--sdk-auth` is not used.
```
jiasli, line 228: There is even a typo!

> **and** not accepted

**and** should be **are**.

README.md
```diff
@@ -225,7 +225,7 @@ The command should output a JSON object similar to this:
...
+- Note: The parameter `--json-auth` outputs the result dictionary accepted by the login action, accessible in Azure CLI versions `>= 2.51.0`. Versions prior to this use `--sdk-auth` with a deprecation warning.
```
jiasli, line 228: Azure CLI uses the word "argument", instead of "parameter":

```suggestion
- Note: The argument `--json-auth` outputs the result dictionary accepted by the login action, accessible in Azure CLI versions `>= 2.51.0`. Versions prior to this use `--sdk-auth` with a deprecation warning.
```

jiasli, line 228: Azure CLI uses the word "argument", instead of "parameter":

```suggestion
- Note: Starting from Azure CLI versions 2.51.0, the argument `--json-auth` outputs the result dictionary accepted by the login action. It is equivalent to `--sdk-auth` in older versions of Azure CLI, but `--sdk-auth` shows deprecation warning.
```

README.md
```diff
@@ -196,16 +196,16 @@ Refer to the [Azure Stack Hub Login Action Tutorial](https://learn.microsoft.com
 
 For using any credentials like Azure Service Principal, Publish Profile etc add them as [secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets) in the GitHub repository and then use them in the workflow.
 
-Follow the following steps to configure Azure Service Principal with a secret:
+Follow the following steps to configure Azure Service Principal with a secret in the scope of `resource-group` as the role of `contributor`:
```
jiasli, line 199:
```suggestion
Follow the following steps to configure Azure Service Principal with a secret at the scope of `{resource-group}` as the role of `contributor`:
```

Jacekey23 (approved):

PR #376, opened 2023-11-27
Azure/login should logout the active account at the beginning

MoChilia:
Azure CLI should run `az logout` to clear the active account at the beginning like what is done in Azure PowerShell: 
https://github.com/Azure/login/blob/34b958dce7c4a4190d4bd0a81255bef97f2e5cf5/src/PowerShell/AzPSScriptBuilder.ts#L27-L28
Otherwise, when the user doesn't explicitly give the `subscription-id` or the given `subscription-id` is invaild, the post steps will continue using the user's active subscription which has set last time and not logged out. 

The following test case will encounter this issue: https://github.com/Azure/login/blob/34b958dce7c4a4190d4bd0a81255bef97f2e5cf5/.github/workflows/azure-login-negative.yml#L355-L366

### DIFF
--- a/src/Cli/AzureCliLogin.ts
+++ b/src/Cli/AzureCliLogin.ts
@@ -34,6 +34,13 @@ export class AzureCliLogin {
         await this.executeAzCliCommand(["--version"], true, execOptions);
         core.debug(`Azure CLI version used:\n${output}`);
 
+        try {
+            await this.executeAzCliCommand(["logout"], true, execOptions);
+        }
+        catch (error) {
+            core.debug(`Ignore logout error: "${error}"`);
+        }
+
         this.setAzurestackEnvIfNecessary();
 
         await this.executeAzCliCommand(["cloud", "set", "-n", this.loginConfig.environment], false);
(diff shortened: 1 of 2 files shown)


YanaXu (approved):

src/Cli/AzureCliLogin.ts
```diff
@@ -34,6 +34,13 @@ export class AzureCliLogin {
         await this.executeAzCliCommand(["--version"], true, execOptions);
         core.debug(`Azure CLI version used:\n${output}`);
 
+        try {
+            await this.executeAzCliCommand(["logout"], true, execOptions);
```
jiasli, line 38: Perhaps `az account clear` for a cleaner environment?

MoChilia: I suppose `az logout` is enough for signing out the current active subscription. And the following steps will use the subscription set in azure/login. There may be a situation that customer needs to switch the active subscription by manually running `az account set`.

MoChilia: In automation, `az account clear` may be a better choice for a cleaner environment. I will modify the code.

PR #385, opened 2023-12-21
Update Workflows

jsoref:
* [Consistently use azure/powershell@v1](https://github.com/Azure/login/commit/2f089bb6435640b3aaf5852edd7537f52b80b50f) -- some were using specific old versions and some were getting the latest -- there was no particular reason for this
* [Consistently use actions/checkout@v4](https://github.com/Azure/login/commit/c4efb1d6ba54c927ba8dc53a7ba0762d661933e2) -- some were using very old (v2), some were using specifically pinned (v3.5.2), and others a floating but outdated version (v3) -- there was no particular reason for this
* [Consistently use actions/setup-node@v4](https://github.com/Azure/login/commit/f9f909a4955f58150045b63d46dc1b2bb3dde8cd) -- some were using the original version (v1) and some the previous version (v3) -- there was no particular reason for this
* [Consistently use actions/github-script@v7](https://github.com/Azure/login/commit/154bfb0d44812e2cde90d89a5bc5bab92741b7f1) -- everyone was using a fairly old version (v3) -- the code here is simple and supported -- there's no reason not to use the current version (old versions will stop working as they rely on old versions of the actions runtime _nodeX_ which are being discontinued incrementally)
* [Consistently use actions/stale@v8](https://github.com/Azure/login/commit/0ef1bf0d96a59bddbf61fc98448872d6b90c742f) -- two out of three were using an old version (v3) while one is actually already using the latest version!

* [Specify codql language as javascript](https://github.com/Azure/login/commit/efdc2b7acd8825373464e38e5fef9c2775ec7d14)  -- CodeQL complains that it can't auto-detect TypeScript, but it will work happily if you tell it that it's JavaScript

* [Name ci.yml](https://github.com/Azure/login/commit/aea517cee534a2a7f54a87552102eff64ea51676) -- Workflows should have names so they look pretty in https://github.com/Azure/login/actions

* [CodeQL does not need a special code](https://github.com/Azure/login/commit/39fdcca1d7e7007698976af935a56e797d2ec1b7) -- CodeQL got smarter and doesn't need special code for checkout

PR #388, opened 2023-12-27
Force failure

jsoref:
Address https://github.com/Azure/login/pull/385#discussion_r1436709259

> As azure/powershell@v1 now supports macOS, we
> need a new way to trigger a failure to show that
> ps steps can fail and their failings can be handled.

I'm not particularly fond of this fix. It depends on what the goal of this step was, and I can't really tell. I looked at the various steps that this workflow had and it seems to have evolved a number of times. The current construct seems pretty complicated for something that's just interested in the exit code of an action (which in turn is dependent on it not supporting the tested platform).

It could be simplified to just `exit 1` instead of retaining the current commands (which previously were never run).


YanaXu:
@jsoref, we're glad you ask this. This workflow file is for negative test cases. It's not for reading but marking all cases we're not supporting right now.
For this case, after your commit was submitted, I've moved all macos test cases to positive test cases. 
I'll close this PR since it's not valid now.

PR #389, opened 2023-12-27
Limit pr-check permissions

jsoref:
Afaict, this workflow does **not** need any interesting [`GITHUB_TOKEN` permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret).

It only needs `contents: read` if it's imported into a private repository (which is something that one might do if one were testing a PR like this).

If it needs `id-token: write`, it should be explicit about that, but based on my reading, it doesn't, as the steps all use `with:`/`creds:`.

Note that as-is, this isn't a security item because this workflow requires approval to run (and given that running it explicitly allows PRs to use an azure credential, I have faith that reviewers are considering the contents of PRs before approving them...).

### DIFF
--- a/.github/workflows/azure-login-pr-check.yml
+++ b/.github/workflows/azure-login-pr-check.yml
@@ -5,6 +5,10 @@ on:
     branches:
       - master
       - 'releases/*'
+
+permissions:
+  contents: read
+
 jobs:
    az-login-test:
      environment: Automation test


.github/workflows/azure-login-pr-check.yml
```diff
@@ -5,6 +5,10 @@ on:
     branches:
       - master
       - 'releases/*'
+
+permissions:
+  contents: read
```
YanaXu, line 10: I still don't understand why to add this line.

jsoref: Try creating a private repository and import this repository, actions/checkout will fail because it requires this permission.

YanaXu: Could you explain how you "import" this repo in your private repo? We do not intend for this workflow to be reused in other repos.

jsoref: 1. visit https://github.com/new/import 
2. enter https://github.com/Azure/login/ as the clone url

I mostly do it if i'm going to be looking for vulnerabilities in a repository...

YanaXu: @jsoref, thanks for sharing this with us. However, the target of this workflow file is only for this repo.  An "imported" repo is a fresh new one with all commits. It's out of our scope.

jsoref:
@YanaXu: the goal of the change is to limit permissions. If you don't want to make the change to `contents: read`, then you should use `permissions: {}`.

Otherwise, any forks (and this workflow isn't restricted to this repository) will run with permissions that far exceed what are necessary.

Issue #426, opened 2024-03-11
Make pre and post action cleanup (`az account clear`) optional for better performance on ephemeral runners


maskati:
The [pre and post action cleanup](https://github.com/Azure/login/blob/81e1d9f360e1ba776e93b8ead1c4320761356b2a/src/cleanup.ts#L7) is somewhat time consuming, especially for otherwise short actions. It is also unnecessary in some contexts, for example when running on GitHub hosted runners or ephemeral self-hosted runners. I agree that the current is a good secure by default implementation, but there should ideally be an opt-out for use cases where it is known that the action will be running in a clean environment with no logged in context, allowing the user of the action to optimize for performance.

MoChilia:
Hi @maskati, this makes sense.  Performing pre/post cleanup for Azure/login on runners that are ephemeral for only one job is unnecessary. We should consider how to bypass such scenarios to save time for users.

kardiojack:
Just wanted to provide support for this improvement. The clearing of credentials is taking over a minute for some of my actions. 60-90 seconds is not a huge deal in isolation, but many actions would take 10-20 seconds _total_ if not for the Az Login pre and post slowdown...and these times add up for a busy repo with a merge queue and many actions run per day.

github-fetch-packages:
It would be really helpful to solve this.

vrdse:
In addition to the performance, the pre login cleanup also throws a warning and is skipped, if pwsh is not yet installed. Thus, I second the request. Though, parameter documentation should be clear about the security implications for non-ephemeral runners.

YanaXu:
Hi @maskati & @Storm-BE & @olafurnielsen & @vrdse & @JosiahSiegel& @lanni-energinet & @kardiojack & @baksetercx , 

we've released a new version [v2.2.0](https://github.com/Azure/login/releases/tag/v2.2.0). In this version, we removed the `pre` step and added 2 env variables to enable/disable pre or post cleanup.
`v2` is aligned with `v2.2.0` now.
Please try the new version. You can also refer to [Enable/Disable the cleanup steps](https://github.com/Azure/login?tab=readme-ov-file#enabledisable-the-cleanup-steps).

Just FYI, the time cost is mainly for login context initialization. After we remove `pre` step, the `main` step may need more time.

Issue #403, opened 2024-01-16
1.6.0 versions Pre login Fails on self hosted runners


kristeey:
Hi,

**Description:**
The 1.6.0 version introduces an error for people using self hosted runners installing `az cli` in a step before using azure/login action in a following step inside the same GHA job. Hence, the new release will do a pre cleanup which uses az cli before the cli is installed -> the job failes. 

**Workaround:** 
Downgrade to 1.5.1. 

**Possible solutions** 
1. Make pre/post cleanup optional. 
2. make az cli install step a part of the action itself.

MoChilia:
For anyone affected by this issue, please first pin the version to `v1.5.1`. We will prepare a fix by skipping pre-cleanup if `az` is not installed. But for security concerns, may I ask about your usage scenario? Are you running `azure/login` in a container? Is it indeed a pure environment with no residual Azure account on your runner?

kristeey:
Our workers are ephemeral and starts up in k8s pods upon request. I.e. The runtime environment are clean Linux containers that are killed when the pipeline finishes.

mattloretitsch-od:
> For anyone affected by this issue, please first pin the version to `v1.5.1`. We will prepare a fix by skipping pre-cleanup if `az` is not installed. But for security concerns, may I ask about your usage scenario? Are you running `azure/login` in a container? Is it indeed a pure environment with no residual Azure account on your runner?

Sure I can fill in that blank.  We are using the Github Runner ARC controller.  It spins fresh runners in k8s with every run.  The image is very basic and has nothing pre-installed on it.

nuryupin-kr:
We install az cli in a step right before calling azure/login step.

ohbriansung:
Please consider rolling back this change and then creating a `v2.0.0` since this is a breaking change for anyone running self-hosted runner without pre-installing the az cli.

YanaXu:
Hi all,

Sorry for introducing this issue. 
**First, plese pin to `v1.5.1` to workaround this issue.**

**What we'll do next:**
Release a new fix in `v1.6.1`. If it's OK, align `v1` to `v1.6.1`.

**What we want to clarify:**

- `v1.6.0` fixes a security issue. It cleans login cache, which is important for regular use scenarios, especially for self-hosted runners on VMs.
[...]

PR #431, opened 2024-04-09
Enhance `pre/post` cleanup by bypassing GitHub-hosted runners

MoChilia:
The purpose of `pre/post` cleanup is to ensure no Azure account remains active before and after a job containing `azure/login`. This measure prevents incorrect operations on unexpected Azure accounts and protects against the disclosure of Azure accounts.

However, certain scenarios are ephemeral for only one job and don't need `pre/post` cleanup. Two main scenarios fall into this category:

1. **GitHub-hosted runners**: A GitHub-hosted runner is on a VM that gets re-imaged for every job. Once the job finishes, the VM is automatically decommissioned (see [Using a GitHub-hosted runner](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#using-a-github-hosted-runner)). Since the VM is ephemeral, we can bypass this scenario to save time for our users. This bypass will be implemented in this PR. To detect whether a runner is GitHub-hosted, I refer to the solution [here](https://github.com/orgs/community/discussions/48359#discussioncomment-5323864).
2. **Running an ephemeral container on a self-hosted runner**: This scenario, as described in (https://github.com/Azure/login/issues/403#issuecomment-1893945688, involves running a container on a self-hosted runner where `az` is not pre-installed. We address this scenario by implementing PR https://github.com/Azure/login/pull/407.

With this PR merged, it is assumed that `pre/post` cleanup will only take effect in scenarios where it is truly required.

- Close https://github.com/Azure/login/issues/426

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -40,7 +40,7 @@ branding:
 runs:
   using: 'node20'
   pre: 'lib/cleanup/index.js'
-  pre-if: ${{ ! startsWith( runner.name, 'GitHub Actions' ) }}
+  pre-if: (! startsWith(runner.name, 'GitHub Actions'))
   main: 'lib/main/index.js'
   post: 'lib/cleanup/index.js'
-  post-if: ${{ ! startsWith( runner.name, 'GitHub Actions' ) }}
+  post-if: (! startsWith(runner.name, 'GitHub Actions'))
--- a/action.yml
+++ b/action.yml
@@ -40,5 +40,7 @@ branding:
 runs:
   using: 'node20'
   pre: 'lib/cleanup/index.js'
+  pre-if: ${{ ! startsWith( runner.name, 'GitHub Actions' ) }}
   main: 'lib/main/index.js'
   post: 'lib/cleanup/index.js'
+  post-if: ${{ ! startsWith( runner.name, 'GitHub Actions' ) }}
--- a/action.yml
+++ b/action.yml
@@ -40,7 +40,7 @@ branding:
 runs:
   using: 'node20'
   pre: 'lib/cleanup/index.js'
-  pre-if: (! startsWith(runner.name, 'GitHub Actions'))
+  pre-if: (! env.AZURE_LOGIN_PRE_CLEANUP_OFF)
   main: 'lib/main/index.js'
   post: 'lib/cleanup/index.js'
-  post-if: (! startsWith(runner.name, 'GitHub Actions'))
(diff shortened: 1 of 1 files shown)


HowardvanRooijen:
Yes please... this would shave a few minutes off every single one of our builds...

maskati:
Thanks for this PR. A note for **2. Running an ephemeral container on a self-hosted runner**, my understanding is that #407 will not fix the case of ephemeral self-hosted runners that already have Azure CLI installed.

It might be quite difficult to determine if a job is running on an ephemeral runner. This information is in the [runner config](https://github.com/actions/runner/blob/1b61d78c07af22165b8f563cf52181e8399668b4/src/Runner.Common/ConfigurationStore.cs#L39), but does not seem to be exposed to the running job. The easiest option would probably be to allow control of this through an action input parameter e.g. `disable-cleanup`.

YanaXu:
Let's take the following questions to consideration:

- Is it a good decision to only bypass GitHub-Hosted runners? 
  - Even if we do descide to only bypass GitHub-Hosted runners, is there a better solution than checking the runner's name? Refer to [runner-context](https://docs.github.com/en/actions/learn-github-actions/contexts#runner-context).
- Can we make it configurable? E.g. an env variable?

kylefossum:
Hi @MoChilia , can we get this patched up and merged? We waste a ton of build time on this.

I agree with Maskati that another input `disable-cleanup` with a default value of `false` would be good.

Then, skipping pre and post steps if `runner.environment ==  github-hosted` OR `disable-cleanup == true`
image

github-fetch-packages:
Would be amazing if we could get this merged!

damianh:
What's the delay here? This is costing is $$$ and is pure wastage.

phj-incom:
Any news here?

HowardvanRooijen:
We got fed up of waiting, forked the repo and reference the fix in our workflows:

```
uses: endjin/login@4ae41681a03e0285d0a85d578968b6ba937c23ee     # forked version including hotfix from https://github.com/Azure/login/pull/431
```

YanaXu:
Hi @HowardvanRooijen, @phj-incom , @damianh , we're sorry for the late response. We're reconsidering the design. If you're blocked by this issue, please consider fork the repo and take the PR change for a workaround, like @HowardvanRooijen did.

YanaXu:
Implement the fix in #484. Close this PR now.

Issue #452, opened 2024-05-29
Azure Government OIDC login default audience fails


jcantosz:
I ran into an issue with the OIDC `audience` used when I used a different `environment`. I expected the default `audience` of the `environment` to be used but instead it was always `api://AzureADTokenExchange`

I would like the action to set the default `audience` based on the `environment` (with a user-specified still value overriding it). If that is not feasible or desirable, I would like the documentation to more explicitly document the need to change the `audience` when not using commercial cloud. 

I would be happy to contribute either of these changes if they would be valuable. For the former, we would need a list of each environment's default audiences 

---
When I was using this to log in to Azure Government with OIDC, I set the following:
```yaml
      - name: Azure login
        uses: azure/login@v2
        with:
          client-id: ${{ secrets.AZURE_CLIENT_ID }}
          tenant-id: ${{ secrets.AZURE_TENANT_ID }}
          subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
          environment: 'AzureUSGovernment'
```
I expected this to work as is, with the `environment` setting up any defaults for Azure Government . The default `audience` when i created my federated id in Azure Government was something like: `api://AzureADTokenExchangeUSGov`, so I had to set that, like below:

```yaml
      - name: Azure login
        uses: azure/login@v2
        with:
          client-id: ${{ secrets.AZURE_CLIENT_ID }}
          tenant-id: ${{ secrets.AZURE_TENANT_ID }}
          subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
          environment: 'AzureUSGovernment'
          audience: api://AzureADTokenExchangeUSGov
```

MoChilia:
Hi @jcantosz, this feature request sounds reasonable, but I couldn't find an official document listing the default exchange token audiences for different clouds. Could you share it with me?

jcantosz:
Thank @MoChilia! I would bet that those default audiences would align to the Azure environments [like those listed here](https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=linux#sample-5-get-the-azure-environment-where-the-vm-is-running) ([from the SDK](https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.management.resourcemanager.fluent.azureenvironment?view=azure-dotnet-legacy)). I will see if I can find confirmation of that

jcantosz:
[...]

PR #455, opened 2024-06-03
Update documentation for setting audience when environment is set

jcantosz:
Updates the documentation to include information about setting the `audience` parameter when the `environment` is not public cloud (default). The `audience` for each cloud environment is different, but not documented on Azure's doc site, so setting it is up to the user.

Closes: #452


MoChilia (approved):
LGTM

PR #471, opened 2024-07-08
Clarify when it's optional to specify the Azure subscription ID

cnotin:
Auth to Entra ID, with this action, does not necessarily need an Azure subscription. For example, if the goal is to access Entra ID / MS Graph only, and not the Azure RM API, it's totally unnecessary.

Related PR in doc: https://github.com/MicrosoftDocs/entra-docs/pull/983

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ Refer to [Login With OpenID Connect (OIDC)](#login-with-openid-connect-oidc-reco
 
 The input parameter `subscription-id` specifies the login subscription id.
 
-It's used in login with OpenID Connect (OIDC) and managed identity.
+It's used in login with OpenID Connect (OIDC), especially for managed identity.
 
 It's better to create a GitHub Action secret for this parameter when using it. Refer to [Using secrets in GitHub Actions](https://docs.github.com/actions/security-guides/using-secrets-in-github-actions).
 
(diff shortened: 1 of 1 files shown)


MoChilia:
Thank you for your contribution! I appreciate your effort and understand your concern, but the statement might not be proper. We may need to refine this section with additional consideration.

cnotin:
Hum what is your concern? How it’s written or what it means? Because I’m very certain that you can login without specifying a subscription if you just want to interact with Entra ID but not Azure resources.

MoChilia:
The `subscription-id` is already marked as an optional parameter, 
image
And a detailed explanation of the `allow-no-subscriptions` parameter with an example is provided in the documentation, https://github.com/Azure/login?tab=readme-ov-file#allow-no-subscriptions.
It is unclear why additional statements are needed, and the use of 'especially for managed identity' seems inappropriate in this context.

MoChilia:
But I like the change about the last part 'Capability has been added to support access to **Entra** tenants without **Azure** subscriptions for both OIDC and non-OIDC. This can be useful to run **Microsoft Entra** tenant level commands, such as `az ad`. The action accepts an optional parameter `allow-no-subscriptions` which is `false` by default.'
Please update the PR if you agree, and we can proceed with merging the change into the document.

cnotin:
> the use of 'especially for managed identity' seems inappropriate in this context.

I changed "and" to "especially" because MIs aren't the only use-case for OIDC. For example, one can authenticate with OIDC with a federated identity bound to a service principal which isn't necessarily an MI

> Before you use Azure Login Action with OIDC, you need to configure a federated identity credential on a service principal or a managed identity. ([source](https://github.com/Azure/login#login-with-openid-connect-oidc-recommended))


Also,
> It is unclear why additional statements are needed

I've added some references to "Entra ID" here and there for folks who want to use this to login to Entra ID only, and who may be confused or discouraged by the references to Azure only.

PR #514, opened 2025-02-20
Use `--client-id` for user-assigned managed identity authentication in Azure CLI v2.69.0 or later.

### LINKED ISSUE #506: Use `--client-id` for user-assigned managed identity authentication
Azure CLI introduced 3 new arguments `--client-id`, `--object-id` and `--resource-id` to replace `--username` for user-assigned managed identity authentication: https://github.com/Azure/azure-cli/pull/30525.

Since `azure/login` only supports `client-id`: https://github.com/marketplace/actions/azure-login#login-with-user-assigned-managed-identity, it should map `client-id` to `--client-id` instead of `--username`.

MoChilia:
Azure CLI introduced 3 new arguments `--client-id`, `--object-id` and `--resource-id` to replace `--username` for user-assigned managed identity authentication: https://github.com/Azure/azure-cli/pull/30525.

Since `azure/login` only supports `client-id`: https://github.com/marketplace/actions/azure-login#login-with-user-assigned-managed-identity, it should map `client-id` to `--client-id` instead of `--username`.

For not introducing breaking changes, `--username` should be kept in Azure/login for Azure CLI versions before 2.69.0. But it is only valid in Azure/login@v2, once Azure CLI removes the argument `--username`, Azure/login@v1 will break for user-assigned managed identity login.

The test workflow: https://github.com/Azure/azclitools-actions-test/blob/main/.github/workflows/azure-login-mi-version-test.yml
Workflow result: https://github.com/Azure/azclitools-actions-test/actions/runs/13513290504

Close #506

### DIFF
--- a/src/Cli/AzureCliLogin.ts
+++ b/src/Cli/AzureCliLogin.ts
@@ -8,6 +8,7 @@ export class AzureCliLogin {
     loginConfig: LoginConfig;
     azPath: string;
     loginOptions: ExecOptions;
+    azVersion: string;
 
     constructor(loginConfig: LoginConfig) {
         this.loginConfig = loginConfig;
@@ -30,6 +31,7 @@ export class AzureCliLogin {
 
         await this.executeAzCliCommand(["version"], true, execOptions);
         core.debug(`Azure CLI version used:\n${output}`);
+        this.azVersion = JSON.parse(output)["azure-cli"];
 
         await this.registerAzurestackEnvIfNecessary();
 
@@ -108,7 +110,14 @@ export class AzureCliLogin {
     }
 
     async loginWithUserAssignedIdentity(args: string[]) {
-        args.push("--username", this.loginConfig.servicePrincipalId);
+        const azcliMinorVersion = parseInt(this.azVersion.split('.')[1], 10);
+        //From Azure-cli v2.69.0, `--username` is replaced with `--client-id`, `--object-id` or `--resource-id`: https://github.com/Azure/azure-cli/pull/30525
+        if (azcliMinorVersion < 69) {
+            args.push("--username", this.loginConfig.servicePrincipalId);
+        }
+        else {
+            args.push("--client-id", this.loginConfig.servicePrincipalId);
+        }
         await this.callCliLogin(args, 'user-assigned managed identity');
     }
 
--- a/src/Cli/AzureCliLogin.ts
+++ b/src/Cli/AzureCliLogin.ts
@@ -31,8 +31,12 @@ export class AzureCliLogin {
 
         await this.executeAzCliCommand(["version"], true, execOptions);
         core.debug(`Azure CLI version used:\n${output}`);
-        this.azVersion = JSON.parse(output)["azure-cli"];
-
+        try {
+            this.azVersion = JSON.parse(output)["azure-cli"];
+        }
+        catch (error) {
+            core.warning("Failed to parse Azure CLI version.");
+        }
         await this.registerAzurestackEnvIfNecessary();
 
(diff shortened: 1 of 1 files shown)


src/Cli/AzureCliLogin.ts
```diff
@@ -30,6 +31,7 @@ export class AzureCliLogin {
 
         await this.executeAzCliCommand(["version"], true, execOptions);
         core.debug(`Azure CLI version used:\n${output}`);
+        this.azVersion = JSON.parse(output)["azure-cli"];
```
YanaXu, line 34: Have you checked all Azure CLI version to see the output format of this? What if `JSON.parse` throws an exception?

src/Cli/AzureCliLogin.ts
```diff
@@ -108,7 +110,14 @@ export class AzureCliLogin {
     }
 
     async loginWithUserAssignedIdentity(args: string[]) {
-        args.push("--username", this.loginConfig.servicePrincipalId);
+        const azcliMinorVersion = parseInt(this.azVersion.split('.')[1], 10);
```
YanaXu, line 113: What if `parseInt` throws an exception? Will it fail the login?

src/Cli/AzureCliLogin.ts
MoChilia: Azure CLI hasn't changed its JSON output format for over 5 years.
image
I prefer to use `az version` to get the Azure CLI version. If `JSON.parse` throws an exception, it means the user is using an unsupported version of Azure CLI and should be rejected.

@jiasli, do you have any other suggestions for reliably obtaining the Azure CLI version?

src/Cli/AzureCliLogin.ts
MoChilia: Not sure how `parseInt` could throw an exception on `azcliMinorVersion`. The Azure CLI version always follows the `xx.xx.xx` format.

src/Cli/AzureCliLogin.ts
jiasli: `az version` is stable. It is safe to use it.

src/Cli/AzureCliLogin.ts
jiasli: The second version section is always an integer.

src/Cli/AzureCliLogin.ts
```diff
@@ -108,7 +110,14 @@ export class AzureCliLogin {
     }
 
     async loginWithUserAssignedIdentity(args: string[]) {
-        args.push("--username", this.loginConfig.servicePrincipalId);
+        const azcliMinorVersion = parseInt(this.azVersion.split('.')[1], 10);
+        //From Azure-cli v2.69.0, `--username` is replaced with `--client-id`, `--object-id` or `--resource-id`: https://github.com/Azure/azure-cli/pull/30525
+        if (azcliMinorVersion < 69) {
+            args.push("--username", this.loginConfig.servicePrincipalId);
+        }
```
jiasli, line 117: Any schedule on when login action stops supporting Azure CLI < 2.69.0?

MoChilia: There are currently no plans to change. Telemetry shows that about 30% of users utilize self-hosted runners, which may not always install the latest Azure CLI. Additionally, all managed identity users are using Azure VMs. So I would like to continue supporting login action across different Azure CLI versions, similar to cli action.

YanaXu (approved):

jiasli (approved):

ragwell (approved):

sins921 (approved):

PR #531, opened 2025-05-23
try this more creative and faster approach to determining azure CLI version

### LINKED ISSUE #527: Azure/login is slow to execute with Hosted Compute Networking
Possibly related to any of these issues (some open, some closed):

https://github.com/Azure/login/issues/456
https://github.com/Azure/login/issues/449
https://github.com/actions/runner-images/issues/10110

--------------------

We're running `Azure/login@v2` on Github-hosted Actions Runners using the Github provided/default Ubuntu 24.04 image.

It seems when using Hosted Compute Networking, the `Azure/login@v2` action takes a long time to complete.

My first theory was "maybe there's an NSG rule doing something weird", but then when I saw just running an `az version` beforehand speeds things up so much, I feel like something else is at play?

Added context:

- for Hosted Compute Networking we're using the default NSG rules listed in here: https://docs.github.com/en/enterprise-cloud@latest/admin/configuring-settings/configuring-private-networking-for-hosted-compute-products/configuring-private-networking-for-github-hosted-runners-in-your-enterprise#prerequisites **The important part: we don't have a "block all Internet" NSG rule**
- we're using `allow-no-subscriptions: true` as this setup is relying on the [Flexible Federated Identity Credentials (Preview)](https://learn.microsoft.com/en-us/entra/workload-id/workload-identities-flexible-federated-identity-credentials?tabs=github) which isn't supported on User-Assigned Managed Identities yet, only App Registrations.

Any additional debugging steps that can be suggested would be appreciated - I can run further tests to try isolate the root cause with maintainer input.

--------------------

Examples:

- just running login, nothing else, 43-49 seconds
  - note the `runs-on`

Image

Image

```
on:
  push:
    branches:
      - main

permissions:
  id-token: write

jobs:
  azure-cli-startup:
    runs-on: ubuntu-latest-2core--azure-vnet
    steps:
      - name: Azure login
        uses: Azure/login@v2
        with:
          allow-no-subscriptions: true
```

- if we run `az version` before a login, that drops to 7 seconds aggregate, 4 seconds for `az version` and 3 seconds for the login 🤯 

Image

```
on:
  push:
    branches:
      - main

permissions:
  id-token: write

jobs:
  azure-cli-startup:
    runs-on: ubuntu-latest-2core--azure-vnet
    steps:
      - name: Run az version
        run: |
          az version

      - name: Azure login
        uses: Azure/login@v2
        with:
          allow-no-subscriptions: true
```

- if we run this without Hosted Compute Networking, its a 5 second execution time
  - note the `runs-on`

Image

```
on:
  push:
    branches:
      - main

permissions:
  id-token: write

jobs:
  azure-cli-startup:
    runs-on: ubuntu-latest
    steps:
      - name: Azure login
        uses: Azure/login@v2
        with:
          allow-no-subscriptions: true
```

CpuID:
An attempt to fix https://github.com/Azure/login/issues/527

### DIFF
--- a/src/Cli/AzureCliLogin.ts
+++ b/src/Cli/AzureCliLogin.ts
@@ -15,21 +15,36 @@ export class AzureCliLogin {
         this.loginOptions = defaultExecOptions();
     }
 
-    async login() {
-        core.info(`Running Azure CLI Login.`);
+    // Normally you would try use "az version" to get this, but the first-time startup time makes it prohibitive
+    // Be more creative...
+    async getAzCliVersion() {
         this.azPath = await io.which("az", true);
         core.debug(`Azure CLI path: ${this.azPath}`);
 
-        let output: string = "";
+        let versionOutput: string = "";
         const execOptions: any = {
             listeners: {
                 stdout: (data: Buffer) => {
-                    output += data.toString();
+                    versionOutput += data.toString();
                 }
             }
         };
 
-        await this.executeAzCliCommand(["version"], true, execOptions);
+        // Due to different installation methods/OSes, take the most compatible approach
+        // Read the 'az' executable content into a string
+        azExecutable = fs.readFileSync(this.azPath, 'utf8');
+        // Modify the script to find the azure-cli package version installed
+        azExecutable = azExecutable.replace(/bin\/python -Im azure.cli.*/g, "bin/pip freeze | grep ^azure-cli== | awk -F== '{print $2}'");
+        // Write the modified script to a temporary file
+        const tempFilePath = path.join(os.tmpdir(), 'az-cli-version.sh');
+        fs.writeFileSync(tempFilePath, azExecutable);
+        // Make the temporary file executable
+        fs.chmodSync(tempFilePath, '755');
+        // Execute the modified script, capture stdout
+        await exec.exec(tempFilePath, [], execOptions);
+        // Remove the temporary file
+        fs.unlinkSync(tempFilePath);
+
         core.debug(`Azure CLI version used:\n${output}`);
         try {
             this.azVersion = JSON.parse(output)["azure-cli"];
@@ -37,6 +52,12 @@ export class AzureCliLogin {
         catch (error) {
             core.warning("Failed to parse Azure CLI version.");
         }
+    }
+
+    async login() {
+        core.info(`Running Azure CLI Login.`);
+
+        await this.getAzCliVersion();
         await this.registerAzurestackEnvIfNecessary();
 
         await this.executeAzCliCommand(["cloud", "set", "-n", this.loginConfig.environment], false);


CpuID:
@microsoft-github-policy-service agree

MoChilia:
Thank you @CpuID for your contribution! I will soon look into it.

CpuID:
No worries @MoChilia - feel free to refactor it to be more idiomatic JS/TS if you want, but hopefully you get the idea of what the goal is :) (to avoid loading *all* Azure CLI command modules unnecessarily, for speed)

src/Cli/AzureCliLogin.ts
```diff
@@ -15,28 +15,49 @@ export class AzureCliLogin {
...
                 }
             }
         };
 
-        await this.executeAzCliCommand(["version"], true, execOptions);
+        // Due to different installation methods/OSes, take the most compatible approach
+        // Read the 'az' executable content into a string
+        azExecutable = fs.readFileSync(this.azPath, 'utf8');
+        // Modify the script to find the azure-cli package version installed
+        azExecutable = azExecutable.replace(/bin\/python -Im azure.cli.*/g, "bin/pip freeze | grep ^azure-cli== | awk -F== '{print $2}'");
```
MoChilia, line 37: This line is not cross-platform. It will work on Unix-like systems but will fail on Windows, since windows doesn't have `grep` and `awk`.

CpuID: I agree good call, if you can't tell I'm not a Windows user ;)

I think if we do that part in typescript, it may be cross-platform? eg. just feed in `azExecutable = azExecutable.replace(/bin\/python -Im azure.cli.*/g, "bin/pip freeze");` then parse the result in typescript?

manender-coder:
az ad sp create-for-rbac --name "github-actions-deploy" --role contributor \
  --scopes /subscriptions/<YOUR_SUBSCRIPTION_ID> \
  --sdk-auth

Hamidsezary (approved):

CpuID (commented):
I wouldn't recommend merging this until the review comment is resolved (I'm not a Windows user so not sure the right way to resolve it)

MoChilia:
@CpuID, thank you for raising this important topic to improve Azure/login performance. As discussed in [Azure/azure-cli#31541](https://github.com/Azure/azure-cli/issues/31541), the main reason for the slow initial run of Azure CLI is the need to generate the command index. The best solution is to pre-warm Azure CLI during runner image creation, as implemented in [actions/runner-images#11173](https://github.com/actions/runner-images/pull/11173). But its approach could potentially disrupt many existing actions.

Would you mind closing this PR? I will look for opportunities to implement this improvement on the runner image side without introducing breaking changes.

Hamidsezary (commented):
Hamidjahanpur@gmail.com

PR #559, opened 2025-10-10
Add support for a custom az-path

duncan3dc:
In some of our workflows we use docker containers to ensure we've got a consistent environment. We use this action to authenticate with OIDC, but this action assumes the `az` CLI is running on the GitHub actions machine.

This PR allows a custom `az` path to be specified, so that we can authenticate inside the docker container, and then run our following `az` commands in there.

For example:
```yaml
- name: Setup docker container
  run: |
    docker build . -t ci-image
    docker run --interactive --detach \
      --env ACTIONS_ID_TOKEN_REQUEST_URL=$ACTIONS_ID_TOKEN_REQUEST_URL \
      --env ACTIONS_ID_TOKEN_REQUEST_TOKEN=$ACTIONS_ID_TOKEN_REQUEST_TOKEN \
      --name ci ci-image

- name: Authenticate with Azure
  uses: azure/login@v2
  with:
    client-id: "${{ secrets.AZURE_CLIENT_ID }}"
    tenant-id: "${{ secrets.AZURE_TENANT_ID }}"
    subscription-id: "${{ secrets.AZURE_SUBSCRIPTION_ID }}"
    az-path: "docker exec ci az"

- name: Now we can run az inside our container
  run: docker exec ci az account show
```

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -34,6 +34,9 @@ inputs:
     description: 'The type of authentication. Supported values are SERVICE_PRINCIPAL, IDENTITY. Default value is SERVICE_PRINCIPAL'
     required: false
     default: 'SERVICE_PRINCIPAL'
+  az-path:
+    description: 'Override the location of the Azure CLI program'
+    required: false
 branding:
   icon: 'login.svg'
   color: 'blue'
--- a/src/Cli/AzureCliLogin.ts
+++ b/src/Cli/AzureCliLogin.ts
@@ -7,6 +7,7 @@ import * as io from '@actions/io';
 export class AzureCliLogin {
     loginConfig: LoginConfig;
     azPath: string;
+    azPathOverridden: boolean;
     loginOptions: ExecOptions;
     azVersion: string;
 
@@ -17,7 +18,8 @@ export class AzureCliLogin {
 
     async login() {
         core.info(`Running Azure CLI Login.`);
-        this.azPath = await io.which("az", true);
+        this.azPath = this.loginConfig.azPath;
+        this.azPathOverridden = this.loginConfig.azPathOverridden;
         core.debug(`Azure CLI path: ${this.azPath}`);
 
         let output: string = "";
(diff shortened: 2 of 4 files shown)


MoChilia:
Hi @duncan3dc, thank you for your contribution. I appreciate this feature. But I would prefer setting it as an environment variable rather than using a parameter. 

@YanaXu  For consistency, should we consider whether Azure PowerShell will also require a custom path?

vinberg88 (approved):

srceenr29y1:
vại tương lơ . Cấm mọi quyền cả admins( nguyên ; Nguyễn ...)
phát hiện ban vĩnh viễn

Vào Thứ 6, 6 thg 3, 2026 lúc 20:06 Mattias Vinberg ***@***.***>
đã viết:

joe345-str (approved):
ok

mrsgoode25-ops (approved):

MaddyMicrosoft:
What this PR seems to address looks achievable two ways today without a new input:

1. Login inside the container directly. Fetch the OIDC token and run the login there with something like: 

```bash
oidc_token=$(curl -s -H "Authorization: bearer $ACTIONS_ID_TOKEN_REQUEST_TOKEN" \
  "$ACTIONS_ID_TOKEN_REQUEST_URL&audience=api://AzureADTokenExchange" | jq -r '.value')
docker exec ci az login --service-principal \
  --username "$AZURE_CLIENT_ID" --tenant "$AZURE_TENANT_ID" --federated-token "$oidc_token"
```

2. Share the credential cache. Set `AZURE_CONFIG_DIR` to a folder bind-mounted into the container, so a normal host login writes there and the container reads it (SDK code can use `AzureCliCredential`). 

Is there a scenario neither of these covers?

On the consistency point: a per-tool path override doesn't scale well. If we add it for the CLI, we'd want an equivalent for Azure PowerShell, and the two store their context differently, so we'd end up maintaining two parallel overrides. Sharing the credential/context location instead covers both `az` and `Connect-AzAccount` uniformly, without a per-tool path.

If there's a real gap, it might be cleanest to capture it as a feature request or PR framed around the shared credential/context location, which works for CLI and PowerShell together.

PR #568, opened 2026-01-23
docs: Update creds structure according to code

nwillems:
small update as I was caught by this one, and read through the code, to find that it is not clienId that was needed.


YanaXu:
“clientId” is correct according to the source code.

PR #600, opened 2026-07-30
Refactor PS login to use a static param()-bound script

MaddyMicrosoft:
## What
Move the Azure PowerShell login step from an interpolated PS script string to a static `AzPSLogin.ps1` file invoked via `pwsh -File`, with user values passed as bound parameters and secret-shaped values passed via env vars.

## Why
The current `AzPSScriptBuilder` concatenates action inputs into a PowerShell script string. Every interpolation is a potential injection surface if escaping is missed for even one field. This refactor gives the PS path the same structural safety the CLI path (`AzureCliLogin.ts`) has always had: **data and code travel through different channels.**

- User values (`tenant-id`, `subscription-id`, `client-id`, `resourceManagerEndpointUrl`) ride as argv → PowerShell parameter binding treats them as inert strings.
- Secret-shaped values (`clientSecret`, `federatedToken`) ride as env vars → they never appear in argv, `ps aux`, or the `core.debug` log.

## Changes
| File | Purpose |
|---|---|
| `src/PowerShell/AzPSLogin.ps1` | New. Static PS script with `param()` binding and splatted `Connect-AzAccount`. |
| `src/PowerShell/AzPSScriptBuilder.ts` | Rewritten. Returns `{ methodName, args, env }` instead of a script string. |
| `src/PowerShell/AzPSLogin.ts` | Calls new invocation API. Debug log now shows argv, not script text. |
| `src/PowerShell/AzPSUtils.ts` | Shared `runPwsh` helper; new `runPSFile(args, env)`. |
| `scripts/copy-ps-assets.js` | New. Copies `AzPSLogin.ps1` into `lib/main/` at build time. |
| `package.json` | `build:main` runs the copy script. |
| `README.md` | Two new `[!WARNING]` blocks: don't pipe `${{ github.event.* }}` into these inputs; keep `enable-AzPSSession` off unless needed. |

## Verification
- **30/30** unit tests pass. New adversarial tests cover single-quote, double-quote, newline, `$`, and switch-shaped values in `tenant`, `subscription`, `client-id`, and `resourceManagerEndpointUrl`.
- End-to-end tested with real `pwsh`: adversarial payloads arrive at `Connect-AzAccount` / `Add-AzEnvironment` as literal string data, not code.
- `npm run build` produces `lib/main/index.js` + `lib/main/AzPSLogin.ps1`.

## Scope notes

**On `LoginConfig` format-validation suggestion:** deliberately not included. Azure's ARM API spec (`Microsoft.Subscription/subscriptions/rename`, `SubscriptionName` schema) allows any Unicode string for subscription display names, and Az.Accounts' `-Subscription` accepts both GUIDs and display names via `[Alias("SubscriptionName", "SubscriptionId")]`. Any character-class check we impose has legitimate false positives against real subscription names like `Prod (US East)`. The refactor closes the exploitation surface structurally; format validation is defense-in-depth we couldn't cleanly bound. 
Happy to add per-field checks for `tenant-id` and `client-id` (which really are always GUID/domain) if reviewers prefer.

## Behaviour change worth noting
`core.debug` log format: previously printed the fully interpolated PS script (including a `ConvertTo-SecureString 'value'` line subject to `core.setSecret` masking). Now prints the argv sent to `pwsh` and contains no secret-shaped values at all. Only visible when `ACTIONS_STEP_DEBUG` is enabled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,16 @@ Azure Login Action supports different ways of authentication with Azure.
 > [!WARNING]
 > Avoid using managed identity login on self-hosted runners in public repositories. Managed identities enable secure authentication with Azure resources and obtain Microsoft Entra ID tokens without the need for explicit credential management. Any user can open pull requests against your repository and access your self-hosted runners without credentials. See more details in [self-hosted runner security](https://docs.github.com/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners#self-hosted-runner-security).
 
+** **
+
+> [!WARNING]
+> Only pass values from `${{ secrets.* }}` into `client-id`, `tenant-id`, `subscription-id`, and `creds`. Do not pipe values from `${{ github.event.* }}` (pull request titles, issue comments, `workflow_dispatch` inputs, branch names, etc.) into these inputs. Untrusted values in these fields can allow attackers to influence the Azure identity the action logs in as.
+
+** **
+
+> [!WARNING]
+> Only set `enable-AzPSSession: true` if your workflow runs Azure PowerShell (`Az.*`) cmdlets. If your workflow only uses the Azure CLI (`az ...`), leave `enable-AzPSSession` unset (the default is `false`). Enabling it launches an additional PowerShell login step that is unnecessary for CLI-only workflows.
+
 ## Input Parameters
 
 |Parameter Name|Required?|Type|Default Value|Description|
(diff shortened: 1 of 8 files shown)


MaddyMicrosoft:
Superseded by #607, which is opened from an in-repo branch (Azure/login) instead of a fork so the credential-gated live login tests can run. Same commits. Continuing there.

PR #614, opened 2026-08-17
Add in dynamic version for telemetry using GITHUB_ACTION_REF

MaddyMicrosoft:
## Summary
Reports the action ref the consumer actually pinned in telemetry, instead of a hardcoded version string. Prevents manual changes and drift. 

## Change
**Telemetry: report the actual action ref (`src/common/Utils.ts`)**
The user-agent string hardcoded `@v3`, which silently goes stale on a major bump. It now reads `GITHUB_ACTION_REF` (the ref the consumer pinned, e.g. `v3`, `v3.1.0`, a SHA, or `main`), falling back to `unknown`. This never goes stale and distinguishes how consumers actually pin.

**Live-tests: isolate the resource group per run**
The workflow used a fixed RG name plus a repo-wide concurrency group. Because the test jobs sit behind the `Automation test` approval gate, an unapproved run held the shared lock in the `waiting` state and blocked live-tests on every other PR. Each run now provisions its own RG (`GitHubAction_CI_Group_${{ github.run_id }}`), and the concurrency group is scoped to the branch, so runs never block across PRs.

**Live-tests: fix RG name expansion on Windows**
The RG name was referenced bash-style (`$RG_POSITIVE`) in `az group` steps. Those use the runner default shell, which is PowerShell on windows-latest, so the value expanded to empty and `az` failed. It now uses the `${{ env.RG_POSITIVE }}` expression, which is shell-agnostic.

## Why
- The version string can no longer go stale on a major bump; it always reflects the real ref.
- It is more accurate telemetry: it distinguishes consumers pinned to a major (`@v3`) from an exact version, a SHA, or `main`.

## Notes
- Behavior change for telemetry consumers: the value after `@` is now the pinned ref rather than always `@v3`, so downstream parsing that assumed `@vN` should bucket the new values. Worth a heads-up to whoever owns the user-agent dashboards.


isra-fel (approved):

isra-fel:
Great idea! Would be nice to have versions even if customer pins a sha or branch, but it's probably not worth it.

PR #641, opened 2026-09-09
Add the ability to prevent the masking of clientId

MaddyMicrosoft:


MaddyMicrosoft:
Created to test https://github.com/Azure/login/pull/634 due to forks not allowing tests with secrets to run.

